PR #134, opened 2019-07-24
More logging

koszti:
If we fast sync lot of tables and one of them fails then we need to wait until the entire process finishing to see exceptions in the summary table.

This PR logs the exception immediately into log file.


koszti:
not required in this format.
same functionality is already in master by https://github.com/transferwise/pipelinewise/commit/d6312c305754bec3017421b8a4960ee29511fd5d

PR #198, opened 2019-09-20
bump target-redshift

koszti:


koszti:
duplicated https://github.com/transferwise/pipelinewise/pull/199

PR #205, opened 2019-09-25
shuffle dev-project files

louis-pie:


koszti:
some files are not part of dev project but required for PROD usage to build and run short living containers. Can you please leave these files in root:
* `Dockerfile`
* `.dockerignore`
* `entrypoint.sh`

Further doc about using docker on Prod:
* https://transferwise.github.io/pipelinewise/installation_guide/installation.html#running-in-docker
* https://github.com/transferwise/pipelinewise#running-from-docker

dev-project/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: python:3.7.4-buster
     container_name: pipelinewise_dev
     working_dir: /usr/src/app
-    entrypoint: /usr/src/app/entrypoint-dev.sh
+    entrypoint: /usr/src/app/dev-project/entrypoint.sh
     environment:
       PIPELINEWISE_HOME: /usr/src/app
```
koszti: I think `PIPELINEWISE_HOME` has to be `/usr/src/app/dev-project` otherwise it will use `.virtualenvs` folder from root and not from `dev-project`.

Changing this might require some other changes in entrypoint-dev.sh

louis-pie: will do it

PR #209, opened 2019-09-26
Adds option to grant user/groups separately in Redshift

koszti:
This PR is for FastSync Redshift compatibility to grant access to users and groups separately. Original issue reported at: https://github.com/transferwise/pipelinewise-target-redshift/issues/15

YAML can be defined in this format as well:
```
schemas:

  - source_schema: "my_db"             # Source schema (aka. database)
    target_schema: "repl_my_db"        # Target schema in the destination Data Warehouse
    target_schema_select_permissions:  # Optional only for Redshift
      users: ["user1", "user2]
      groups: ["group1", "group2"]
```

This is only for AWS Redshift. Other supported target databases (snowflake and postgres) can still use the list of roles as earlier.

### DIFF
--- a/pipelinewise/fastsync/commons/target_postgres.py
+++ b/pipelinewise/fastsync/commons/target_postgres.py
@@ -96,7 +96,10 @@ class FastSyncTargetPostgres:
         #self.s3.delete_object(Bucket=bucket, Key=s3_key)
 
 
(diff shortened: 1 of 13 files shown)


koszti:
added more tests and sync with https://github.com/transferwise/pipelinewise-target-postgres/blob/master/tests/integration/test_target_postgres.py#L430

PR #210, opened 2019-09-26
Update mysql.rst

bparlapalliWiley:
while working on MySQL tap - realised I needed access to Information_schema tables as well. this was not explicitly mentioned in the documentation.


koszti:
thanks for the contribution

Issue #203, opened 2019-09-24
How to use environment variables for secrets in the YAML


vitorbaptista:
Is it possible to use env variables in the YAML files for the secrets, instead of using Ansible Vault? If not, is it a feature you'd consider adding?

koszti:
not possible at the moment, but if it looks beneficial for us, we can consider it. Alternatively feel free to send PRs if you feel like developing. We can review PRs quickly.

sokser:
I have a PR that will allow them to be used in the singer spec that you use: https://github.com/transferwise/pipelinewise/pull/223

johnoscott:
As an example of how this might work, the DBT project supports environment variables in YAML configs like this : 

```yaml
profile:
  target: prod
  outputs:
    prod:
      type: postgres
      host: 127.0.0.1
      # IMPORTANT: Make sure to quote the entire Jinja string here
      user: "{{ env_var('DBT_USER') }}"
      pass: "{{ env_var('DBT_PASSWORD') }}"
      ....
```

https://docs.getdbt.com/docs/env_var

johnoscott:
and (of course) Docker supports environment variables in Dockerfile and docker-compose.yaml

https://docs.docker.com/compose/environment-variables/

```version: "3"

services:
  console:
    image: ${ECR_REGISTRY_ID:?err}.dkr.ecr.${REGION:?err}.amazonaws.com/${REPO_CONSOLE:?err}:latest
    volumes:
      - data:/shared
      - /var/run/docker.sock:/var/run/docker.sock #enable docker commands to host
    links:
      - runner
    command: /app/entrypoint.sh
    logging:
      driver: awslogs
      options:
        awslogs-group: ${PROJECT_PREFIX:?err}
        awslogs-region: ${REGION:?err}
        awslogs-stream-prefix: console
  runner:
    image: ${ECR_REGISTRY_ID:?err}.dkr.ecr.${REGION:?err}.amazonaws.com/${REPO_RUNNER:?err}:latest
    volumes:
      - data:/shared
    logging:
      driver: awslogs
      options:
        awslogs-group: ${PROJECT_PREFIX:?err}
        awslogs-region: ${REGION:?err}
        awslogs-stream-prefix: runner
volumes:
  data:```

Samira-El:
Hey,
This should be possible now thanks to https://github.com/transferwise/pipelinewise/pull/588.

PR #223, opened 2019-10-03
Added the ability to use environmental variables in json files

sokser:
This uses Jinja2 to preparse config files that are loaded into JSON. It attempts to replace `{{ VARIABLE }}` with the corresponding environmental variable.

### DIFF
--- a/pipelinewise/cli/utils.py
+++ b/pipelinewise/cli/utils.py
@@ -24,6 +24,9 @@ from ansible.module_utils._text import to_text
 from ansible.module_utils.common._collections_compat import Mapping
 from ansible.errors import AnsibleError
 
+from jinja2 import Template
+import os
+
 from . import tap_properties
 
 logger = logging.getLogger('Pipelinewise CLI')
@@ -92,7 +95,8 @@ def load_json(path):
         logger.debug('Parsing file at {}'.format(path))
         if os.path.isfile(path):
             with open(path) as f:
-                return json.load(f)
+                template = Template(f.read())
+                return json.load(template.render(**os.getenv))
         else:
             logger.debug("No file at {}".format(path))
             return None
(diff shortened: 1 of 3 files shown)


vitorbaptista:
Looking forward for this to be merged!

Personally, I'd rather have something like `{{ env['MY_ENV'] }}` instead of `{{ MY_ENV }}`. This should be simply changing `template.render(**os.getenv)` to `template.render(env=os.environ)` (maybe `os.environ.copy()` to avoid passing the actual env dict)

koszti:
I think dynamically rendering the JSON files will have some problems because in this case every tap and target connector should do the same thing well and I think that's not the case in the connectors and they will raise Invalid JSON example at runtime. Remember, this pipelinewise library is only a "wrapper" on top of other singer connectors. It's generating JSON config files for singer connectors from a common YAML structure and running the singer taps/targets individually.

Instead of rendering the JSON we can try to template the YAML files with similar methods. But please also note if we will use env variables in the YAML (and this was the original question #203), then it will be evaluated only when importing the project by `pipelinewise importing --dir ..`! (the time when generating singer compatible JSON files from the common pipelinewise YAML)


If you think that evaluating environment variables only when importing the project then it's a very similiar change to this PR but in the YAML at around https://github.com/transferwise/pipelinewise/blob/master/pipelinewise/cli/utils.py#L157

What do you think?

johnoscott:
I agree dynamic JSON is not appropriate for PipelineWise since it is the output of the YAML config.

I do support dynamic YAML with environment variables for USE CASES such as supporting different deployment stages (DEV, TEST, PROD etc).  [DBT](https://www.getdbt.com/) comes to mind as an example project that does this well I think.

guy-adams:
Does this need a documentation update too?

Samira-El:
Hey,
This should be possible now thanks to #588.

johnoscott:
Awesome, awesome, awesome !

PR #229, opened 2019-10-04
update target_snowflake.py

### LINKED ISSUE #228: Target-snowflake does not create schema and tables with key words
In target snowflake, we cannot create tables or schemas with keyword names
for example:
`Schema name AS failed (created schemas AS AS)`
`Table name ORDER failed (Create table ORDER - ORDER is reserved key word in Snowflake)`

there is a workaround for this and we have got it working - but needs to be tested fully with any other issues.

bparlapalliWiley:
allow creation of table columns with reserved key words.
Snowflake has reserverd key words (ORDERS, AS etc full list here: [https://docs.snowflake.net/manuals/sql-reference/reserved-keywords.html])

some key words like "ORDER" are frequently used as column or table names.

update the create table SQL to allow for creation of columns with keywords.
does not allow schemas and tables with keywords but a similar fix can be used if needed

this commit partially fixes #228

### DIFF
--- a/pipelinewise/fastsync/commons/target_snowflake.py
+++ b/pipelinewise/fastsync/commons/target_snowflake.py
@@ -103,14 +103,14 @@ class FastSyncTargetSnowflake:
         target_table = table_dict.get('table_name') if not is_temporary else table_dict.get('temp_table_name')
 
         if primary_key:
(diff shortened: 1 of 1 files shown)


Samira-El:
Fixed as of [v0.13.2](https://github.com/transferwise/pipelinewise/tree/v0.13.2).

Issue #242, opened 2019-10-22
MySQL taps white space in either table or column names breaks


guy-adams:
Both of these are allowed in ANSI SQL.

To fix all MySQL queries need to have the table and column names correctly quoted. Similarly the writes into the target need to be similarly quoted. This should also fix #228 since the quoting required for allowing columns with names that are Reserved words (like Group).

I plan to issue a pull request for the fix for this.

koszti:
@guy-adams please be careful with quoting column names, especially on the target-snowflake side. We want to keep the not case sensitive SQL behaviour in every target so all the below SQL should give the same result:
`SELECT columnone FROM myschame.MYTABLE`
`SELECT columnOne FROM myschame.myTable`
`SELECT COLUMNone FROM myschame.MyTaBlE`

If we quote to the table and column names in target then probably we need to use it in every SQL (`SELECT * FROM myschema."myTable"`). This is not a great experience when writing queries in the DWH.

Related to https://github.com/transferwise/pipelinewise/pull/243

guy-adams:
@koszti - that makes sense - and in fact with this pull request in place I actually managed to break something of my own by breaking a query to snowflake that used to work due to forcing case. It may be the way that I have solved it is not the right way. None the less I believe pipelinewise should be able to handle source tables and columns with whitespace. The other option is to simply remove the whitespace and replacing with a "-" or "_" or just removing entirely and having the words run together? This creates slightly undesirable behavious where the target tables have different names to the source tables, but this is still better than not ingesting them at all. What are your thoughts?

koszti:
Yes, I agree, I think replacing whitespaces to "-" or "_" makes sense. Furthermore that doesn't need quoted object names and should not break the current query behaviour in the targets.

Btw, using `_` (underscore) and not `-` is maybe better. Minus signs in object names I think needs to be quoted but underscores are fine.

MihalisW:
In the case of reserved keywords how about dealing with them by appending an `_` (underscore), rather than quoting them?

koszti:
you mean adding an `_` (underscore) prefix or postfix to the colum names? i.e. `SELECT order_ FROM mytable` ?

MihalisW:
Yep, that's correct.

guy-adams:
The only issue I see with this is that it requires us to define an explict list of reserved words rather than "just escape everything" - this may not be an issue, it just gives something that needs to be maintained per target

koszti:
[...]

PR #243, opened 2019-10-22
Adding quoting to all MySql fastsync tap and Snowflake target to allo…

guy-adams:
…w source table and column names with swhitespace and target table and column names with either whitespace, or using reserved works (like 'group')

If there are things I need to do to improve this pull request please let me know. I've tested this against multiple mysql sources with spaces in table and column names and snowflake columns with reserved names.

### DIFF
--- a/pipelinewise/fastsync/commons/tap_mysql.py
+++ b/pipelinewise/fastsync/commons/tap_mysql.py
@@ -106,7 +106,8 @@ class FastSyncTapMySql:
 
 
     def get_primary_key(self, table_name):
-        sql = "SHOW KEYS FROM {} WHERE Key_name = 'PRIMARY'".format(table_name)
+
+        sql = "SHOW KEYS FROM {} WHERE Key_name = 'PRIMARY'".format(utils.backtick_quote_dbtablename(table_name))
         pk = self.query(sql)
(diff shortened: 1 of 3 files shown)


koszti:
@guy-adams we should have equivalent changes https://github.com/transferwise/pipelinewise-target-snowflake repo to keep fastsync and singer functionalities in sync.

Do you have any PR that addresses the same issue in target-snowflake in the above repo?

guy-adams:
No, but I can add it - leave it with me!

koszti:
@guy-adams this is great but please be careful with quoting column names, especially on the target-snowflake side. We want to keep the not case sensitive SQL behaviour in every target so all the below SQL should give the same result:
SELECT columnone FROM myschema.MYTABLE
SELECT columnOne FROM mySchema.myTable
SELECT COLUMNone FROM MYSCHEMA.MyTaBlE

If we quote to the table and column names in target then probably we need to use it in every SQL (SELECT * FROM myschema."myTable") in the target DB as well. This is not a great experience when writing queries and we'd like to avoid it.

Related to https://github.com/transferwise/pipelinewise/issues/242

guy-adams:
Dropping this pull request to handle a different way

Issue #222, opened 2019-10-03
Show the tap-run log output to std-out (so i dont need a second terminal to view the logs)


johnoscott:
I use a second terminal to tail the tap-run logs using multitail.

```
# TIP: tail the logs in a second terminal with multitail
cd ~/.pipelinewise/<project>/<target>/log/
multitail -Q 1 ./*.running
```

... but I would prefer to have the log output to my console.

koszti:
fixed by #245 . Now you can use the `--extra_log` optional argument:

```
pipelinewise run_tap --tap <tap_id> --target <target_id> --extra_log
```

PR #245, opened 2019-10-23
Add extra_log parameter to CLI

fabien-sarcel:
Extra option to get tap-run and fastsync log output into std-out (#222 ).

I run PipelineWise Docker with Airflow (_DockerOperator_) and I need this option to see all the logs into Airflow.

### DIFF
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -35,6 +35,7 @@ name_help = """Name of the project"""
 secret_help = """Path to vault password file"""
 version_help = """Displays the installed versions"""
 log_help = """File to log into"""
+extra_log_help = """Copy singer and fastsync logging into PipelineWise logger"""
 debug_help = """Forces the debug mode with logging on stdout and log level debug."""
 
 def main():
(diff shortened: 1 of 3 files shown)


guy-adams:
Thanks for this - this will be really useful for us running in Docker too!

koszti:
@fabien-sarcel this is great. can you please add a line into the documentation about extra_log parameter at https://transferwise.github.io/pipelinewise/user_guide/cli.html ?
The editable RST format is at https://github.com/transferwise/pipelinewise/blob/master/docs/user_guide/cli.rst

thanks

ivan-transferwise:
Looks good 👍

fabien-sarcel:
I looked at the documentation and I'm not very comfortable with it. I haven't work with Sphinx yet, but above all I'm not sure where I should put the extra_log option (there is no mention of existing log and debug options...).

Moreover, in the HTML documentation, options are rendered with 1 hyphen instead of 2.
I tried to change it without success, so I looked at the Airflow documentation (https://airflow.apache.org/cli.html#Sub-commands:)
and I saw that those command options are automatically generated.
Maybe that's something we should look into ?

koszti:
@fabien-sarcel no worries I added it to the doc page and now it got merged. Thank you for contributing.

Issue #192, opened 2019-09-12
Failing to install from source on MacOS


rickybasse:
I am having toubles installing the tool from source following the [guidelines](https://transferwise.github.io/pipelinewise/installation_guide/installation.html#installing-from-source)

### Expected Behavior

The tool should complete the steps and successfully install 

### Actual Behavior

Executing `./install.sh --acceptlicenses` results in: 
```sh
...
Collecting PyMySQL==0.7.11 (from pipelinewise==0.10.1)
  Using cached https://files.pythonhosted.org/packages/c6/42/c54c280d8418039bd2f61284f99cb6d9e0eae80383fc72ceb6eac67855fe/PyMySQL-0.7.11-py2.py3-none-any.whl
Collecting psycopg2==2.8.2 (from pipelinewise==0.10.1)
  Using cached https://files.pythonhosted.org/packages/23/7e/93c325482c328619870b6cd09370f6dbe1148283daca65115cd63642e60f/psycopg2-2.8.2.tar.gz
    ERROR: Command errored out with exit status 1:
     command: /Users/rbassetto/babbel/repos/pipelinewise/.virtualenvs/pipelinewise/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/sh/gf_0zlrs6pl42pd9jg3yj8n80000gn/T/pip-install-vt2pyf42/psycopg2/setup.py'"'"'; __file__='"'"'/private/var/folders/sh/gf_0zlrs6pl42pd9jg3yj8n80000gn/T/pip-install-vt2pyf42/psycopg2/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: /private/var/folders/sh/gf_0zlrs6pl42pd9jg3yj8n80000gn/T/pip-install-vt2pyf42/psycopg2/
    Complete output (23 lines):
    running egg_info
    creating pip-egg-info/psycopg2.egg-info
    writing pip-egg-info/psycopg2.egg-info/PKG-INFO
    writing dependency_links to pip-egg-info/psycopg2.egg-info/dependency_links.txt
    writing top-level names to pip-egg-info/psycopg2.egg-info/top_level.txt
[...]

PR #247, opened 2019-10-23
"pg_config executable not found" while running install.sh

ivan-transferwise:
Some discussion here: https://stackoverflow.com/questions/21079820/how-to-find-pg-config-path

Screenshot 2019-10-23 at 15 59 38


koszti:
@ivan-transferwise is this similar to https://github.com/transferwise/pipelinewise/issues/192 ?

PR #248, opened 2019-10-24
AP-207: PipelineWise - Validating YAML config files from CLI

ivan-transferwise:
Along the way also fixed loading schema every time we load tap or target.

### DIFF
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -90,7 +90,7 @@ def main():
         # python keyword and can't be used as function name
         args.command = 'import_project'
 
-    if args.command == 'validate_config':
+    if args.command == 'validate':
         if args.dir == '*':
             print("You must specify a directory path with config YAML files using the argument --dir")
             sys.exit(1)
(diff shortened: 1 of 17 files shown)


koszti:
can we document it at https://transferwise.github.io/pipelinewise/user_guide/cli.html ?
editable RST format is at https://github.com/transferwise/pipelinewise/blob/master/docs/user_guide/cli.rst

pipelinewise/cli/__init__.py
```diff
@@ -82,13 +83,18 @@ def main():
...
         if args.dir == '*':
-            print("You must specify a directory path with config YAML files using the argumant --dir")
+            print("You must specify a directory path with config YAML files using the argument --dir")
             sys.exit(1)
 
         # Every command argument is mapped to a python function with the same name, but 'import' is a
         # python keyword and can't be used as function name
         args.command = 'import_project'
 
+    if args.command == 'validate_config':
```
koszti: can we change it to `args.command == 'validate'` to be in sync to the `commands` list?

Currently the `args.dir` check is not running when running without the required `--dir` arg:
(pipelinewise) root@be2ea33b59aa:/opt/pipelinewise# pipelinewise validate 
2019-10-29 14:56:03 INFO: Searching YAML config files in *
...FileNotFoundError: [Errno 2] No such file or directory: '*'

ivan-transferwise: @koszti right, sorry, totally forgot to change it!

PR #249, opened 2019-10-25
Change format-strings to f-strings

kasparg:
PipelineWise does lots of string formatting and logging. There are some low hanging fruits for making replication jobs faster by optimizing those string operations. 

Simple tests show f-strings are 2x faster than .format(). 

This PR does not cover the whole codebase.

### DIFF
--- a/pipelinewise/cli/utils.py
+++ b/pipelinewise/cli/utils.py
@@ -89,15 +89,15 @@ def load_json(path):
     Deserialise JSON file to python object
     '''
     try:
-        logger.debug('Parsing file at {}'.format(path))
+        logger.debug(f'Parsing file at {path}')
         if os.path.isfile(path):
             with open(path) as f:
                 return json.load(f)
         else:
-            logger.debug("No file at {}".format(path))
+            logger.debug(f"No file at {path}")
             return None
     except Exception as exc:
-        raise Exception("Error parsing {} {}".format(path, exc))
+        raise Exception(f"Error parsing {path} {exc}")
 
 
 def is_state_message(line: str) -> bool:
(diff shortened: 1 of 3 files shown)


pipelinewise/cli/utils.py
```diff
@@ -426,25 +426,25 @@ def run_command(command, log_file=None, line_callback=None):
...
-        log_file_failed = "{}.failed".format(log_file)
-        log_file_success = "{}.success".format(log_file)
+        log_file_running = f"{log_file}.running"
+        log_file_failed = f"{log_file}.failed"
+        log_file_success = f"{log_file}.success"
 
         # Start command
         proc = Popen(shlex.split(piped_command), stdout=PIPE, stderr=STDOUT)
-        f = open("{}".format(log_file_running), "w+")
+        f = open(f"{log_file_running}", "w+")
```
ivan-transferwise: Why to use f-string here? `log_file_running` seems already to be a string.

pipelinewise/cli/utils.py
```diff
@@ -383,9 +383,9 @@ def get_tap_stream_id(tap, database_name, schema_name, table_name):
     pattern = get_tap_property(tap, 'tap_stream_id_pattern')
 
     return pattern \
-        .replace("{{database_name}}", "{}".format(database_name)) \
-        .replace("{{schema_name}}", "{}".format(schema_name)) \
-        .replace("{{table_name}}", "{}".format(table_name))
+        .replace("{{database_name}}", f"{database_name}") \
```
ivan-transferwise: Same here, why not just database_name?

ivan-transferwise:
Left some small comments, otherwise looks good 👍

PR #260, opened 2019-11-04
Update sf

Samira-El:
Upgrading snowflake connector lib and target snowflake to avoid having problem with azure dependency.


koszti:
closing this due to some dependency issues caused by the latest snowflake-python-connector (2.0.3). Same changeset and the fix is in #262

PR #273, opened 2019-11-25
FastSync tap s3 csv to snowflake

Samira-El:
https://transferwise.atlassian.net/browse/AP-318

Implementation of fast sync from s3 csv to snowflake to support full table replication and fix the greenhouse tap.


ivan-transferwise:
There seems to be also a lot of refactoring in your PR. Is that correct? Like refactoring from double brackets to single brackets ('' -> ""). I think it would be a good idea to first setup pylint and all the code rules that we agree on (e.g. using single brackets instead of double ones, etc). wdyt?

Samira-El:
> There seems to be also a lot of refactoring in your PR. Is that correct? Like refactoring from double brackets to single brackets ('' -> ""). I think it would be a good idea to first setup pylint and all the code rules that we agree on (e.g. using single brackets instead of double ones, etc). wdyt?

If i recall those were to fix the docs part of methods, where triple double quote are what needs to be used. But yeah, we definitely need pylint. I use PyCharm with pylint so I fix the code as I write/encounter it.

PR #275, opened 2019-11-25
Selectively install singer connectors

koszti:
This PR adding the option to install only specific singer connectors from the command line.

Available options:
* `./install.sh` : Installs only the default singer connectors. `DEFAULT` and `EXTRA` connectors specified here at: https://github.com/transferwise/pipelinewise/blob/0662cf0f31703816d37c8d03e44daedb4ef7ae6b/install.sh#L160
* `./install.sh --connectors=all` : Finds and installs every available singer connectors automatically
* `./install.sh --connectors=tap-mysql,tap-postgres,target-snowflake` : Installs only a specific list of singer connectors

* `./install.sh --clean` : Clean and delete every previously installed component from the system

### DIFF
--- a/docs/installation_guide/installation.rst
+++ b/docs/installation_guide/installation.rst
@@ -57,10 +57,10 @@ executable docker image:
     $ docker build -t pipelinewise:latest .
 
 
-Building the image may take 5-10 minutes depending on your network connection. At the moment
-there is no official, pre-built image available to download on DockerHub. Once the image is
-ready, create an alias to the docker wrapper script so you can use the ``pipelinewise``
-executable commands everywhere on your system:
+Building the image may take 5-10 minutes depending on your network connection. The output image will
+contain every supporter singer connectors. At the moment there is no official, pre-built image available
+to download on DockerHub. Once the image is ready, create an alias to the docker wrapper script so you can
+use the ``pipelinewise`` executable commands everywhere on your system:
 
 .. code-block:: bash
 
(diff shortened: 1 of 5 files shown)


docs/installation_guide/installation.rst
```diff
@@ -120,6 +120,60 @@ the installation and accept every license agreement run ``./install --acceptlice
...
+    $ ./install --connectors=tap-mysql,tap-postgres,target-snowflake
+
+Here’s the list of the singer connectors and if they are installed by default or not:
+
++----------------------------+-----------------------------------------+----------------------------------+---------------------------------------+
+| **Connector**              | **Install Command**                     | **Included in default install?** | **Note**                              |
++----------------------------+-----------------------------------------+----------------------------------+---------------------------------------+
+| all                        | ./install --connectors=all              |                                  | Installs every supported connector    |
++----------------------------+-----------------------------------------+----------------------------------+---------------------------------------+
+| tap-adwords                | ./install --connectors=tap-adwords      |                                  |                                       |
```
ivan-transferwise: Why there is no value for some taps in column "Included in default install?" ?

ivan-transferwise: Should it be "NO"? :)

koszti: I left it empty for better visibility but it's not trivial so fixed it by 89f08c7ecd5a6de1bec5cd27e3d0ce5d4f91507e

thanks

PR #283, opened 2019-12-03
Few improvs

Samira-El:
* Support date_overrides given in the configuration when doing FastSync.
* refactored type guessing a bit - no change in behavior.

### DIFF
--- a/tests/units/fastsync/test_fastsync_tap_s3_csv.py
+++ b/tests/units/fastsync/test_fastsync_tap_s3_csv.py
@@ -36,10 +36,10 @@ class TestFastSyncTapS3Csv(TestCase):
 
         def tap_type_to_target_type(tap_type):
             return {
-                'Bool': 'smallint',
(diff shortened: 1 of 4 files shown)


tests/units/fastsync/test_fastsync_tap_s3_csv.py
```diff
@@ -36,10 +36,10 @@ def setUp(self) -> None:
 
         def tap_type_to_target_type(tap_type):
             return {
-                'Bool': 'smallint',
-                'Integer': 'number',
-                'Decimal': 'number',
-                'String': 'varchar'
+                'boolean': 'boolean',
```
louis-pie: Should this not be varchar also ?

Samira-El: That's just a mocked function. it can be anything you want. It's injected when creating a class instance.

PR #290, opened 2019-12-27
Fix for transformation type Hidden - mysql

SaurabhJain2611:
Applied fix of hidden mask transformation issue - mysql


koszti:
thanks for this bugfix @SaurabhJain2611 . This is definitely something that we need to consider to merge. Unfortunately we don't merge PRs at this time of the year so still need to wait a few more days but will do in the very beginning of the new year.

PR #292, opened 2019-12-31
fixed issue when missing required PK not detected by post_import_tap_checks

koszti:
**Description**
If `primary_key_required` key not defined explicitly in the target YAML then `post_import_tap_checks` not reporting that source table has no PK.

**Reason**
`primary_key_required` is an optional parameter in the target YAMLs and defaults to `True` but `_run_post_import_tap_checks` expecting the default as `False`. This PR makes it to be in sync and adding some new tests to cover the scenario when `primary_key_required` is not explicitly defined in the target YAML.


Samira-El:
Looks good.

PR #293, opened 2020-01-03
support binary and varbinary columns in Mysql to SF fastsync

Samira-El:
Supporting binary and varbinary columns in Mysql sources when using fast sync.

Atm, I'm fetching the right pipelinewise-tap-mysql and pipelinewise-target-snowflake using git urls because the latest changes are not yet released on PyPi. I will update this PR once those changes are released, so this PR shouldn't be merged until then.

Last change to finish AP-111.

TODO before merge:
1- Update requirements files of pipelinewise-tap-mysql & pipelinewise-target-snowflake connectors.
2- bump pipelinewise minor

### DIFF
--- a/pipelinewise/fastsync/commons/tap_mysql.py
+++ b/pipelinewise/fastsync/commons/tap_mysql.py
@@ -1,3 +1,5 @@
+import codecs
+
 import pymysql
 import gzip
 import csv
@@ -144,8 +146,10 @@ class FastSyncTapMySql:
                             data_type,
                             column_type,
                             CASE
-                            WHEN data_type IN ('blob', 'tinyblob', 'mediumblob', 'longblob', 'binary', 'varbinary', 'geometry')
+                            WHEN data_type IN ('blob', 'tinyblob', 'mediumblob', 'longblob', 'geometry')
                                     THEN concat('hex(', column_name, ')')
+                            WHEN data_type IN ('binary', 'varbinary')
+                                    THEN COLUMN_NAME
                             WHEN data_type IN ('bit')
                                     THEN concat('cast(`', column_name, '` AS unsigned)')
                             WHEN data_type IN ('datetime', 'timestamp', 'date')
@@ -165,10 +169,11 @@ class FastSyncTapMySql:
             """.format(table_dict.get('schema_name'), table_dict.get('table_name'))
         return self.query(sql)
 
-
     def map_column_types_to_target(self, table_name):
         mysql_columns = self.get_table_columns(table_name)
-        mapped_columns = ["{} {}".format(pc.get('column_name'), self.tap_type_to_target_type(pc.get('data_type'), pc.get('column_type'))) for pc in mysql_columns]
+        mapped_columns = ["{} {}".format(pc.get('column_name'),
+                                         self.tap_type_to_target_type(pc.get('data_type'), pc.get('column_type'))) for
+                          pc in mysql_columns]
 
         return {
             "columns": mapped_columns,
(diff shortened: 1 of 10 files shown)


pipelinewise/fastsync/commons/tap_mysql.py
```diff
@@ -207,15 +212,37 @@ def copy_table(self, table_name, path):
...
+                                # ref: https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html
+                                #
+                                # Leaving the trailing zeroes makes FT & Incremental behave differently than Binlog
+                                # and we end up with inconsistent state records.
+                                #
+                                # Example: for a binary value b'pk0'
+                                #   - Binlog would read it as b'pk0' then in HEX it would be b'706b30'
+                                #   - FT & Inc would read it as b'pk0\x00\x00\x00\x00....\x00' then in HEX
+                                #       it would be b'706b300000...0'
+                                stripped_utf8_elem = val.decode().rstrip('\x00').encode()
```
ivan-transferwise: The same code as in https://github.com/transferwise/pipelinewise-tap-mysql/pull/9
So same concerns apply here :)

pipelinewise/fastsync/commons/tap_mysql.py
```diff
@@ -207,15 +212,37 @@ def copy_table(self, table_name, path):
...
-                        utils.log("Exporting batch from {} to {} rows from {}...".format((exported_rows-export_batch_rows),exported_rows, table_name))
+                    if len(rows) == export_batch_rows:
+                        # Then we believe this to be just an interim batch and not the final one so report on progress
+                        utils.log(
+                            "Exporting batch from {} to {} rows from {}...".format((exported_rows - export_batch_rows),
+                                                                                   exported_rows, table_name))
 
                     # Write rows to file
                     for row in rows:
+                        row = list(row)
```
ivan-transferwise: If you just need to transform every element of the `row`, you can also simplify this code to one line by using `map()`

singer-connectors/tap-mysql/requirements.txt
```diff
@@ -1 +1 @@
-pipelinewise-tap-mysql==1.0.7
+-e git+https://github.com/transferwise/pipelinewise-tap-mysql.git@support-binary-cols#egg=pipelinewise-tap-mysql
```
ivan-transferwise: This is only for tests, right? In the final version you can put `1.1.1` here.

singer-connectors/target-snowflake/requirements.txt
```diff
@@ -1 +1 @@
-pipelinewise-target-snowflake==1.1.8
+-e git+https://github.com/transferwise/pipelinewise-target-snowflake.git@AP-111#egg=pipelinewise-target-snowflake
```
ivan-transferwise: same

pipelinewise/fastsync/commons/tap_mysql.py
koszti: is this the same functionality to the previously used mysql `HEX(...)` function that we do at export time at https://github.com/transferwise/pipelinewise/pull/293/files/a76d4e4609350f89b44a572c202c9d37b97dfe6b#diff-8086c1c0711f0b49a47a6c60db4e2260R150?

If yes, maybe we can leave it to MySQL because these operations usually faster doing in mysql than doing it by python.

Samira-El: > The same code as in [transferwise/pipelinewise-tap-mysql#9](https://github.com/transferwise/pipelinewise-tap-mysql/pull/9)
> So same concerns apply here :)

Check my comments in https://github.com/transferwise/pipelinewise-tap-mysql/pull/9



> is this the same functionality to the previously used mysql `HEX(...)` function that we do at export time [here](https://github.com/transferwise/pipelinewise/pull/293/files/a76d4e4609350f89b44a572c202c9d37b97dfe6b#diff-8086c1c0711f0b49a47a6c60db4e2260R149) ?
> 
> If yes, maybe we can leave it to MySQL because these operations usually faster doing in mysql than doing it by python.

I have thought of that, Hex is not enough to trim the extra 0s, I have spent 2-3 days looking for ways to trim using sql but nothing worked. So doing it using code was my last option. If you know how to do it please do share.

pipelinewise/fastsync/commons/tap_mysql.py
Samira-El: Not every element, just those of type bytes.

singer-connectors/tap-mysql/requirements.txt
Samira-El: I'll update this once the singer components are released.

singer-connectors/target-snowflake/requirements.txt
Samira-El: https://github.com/transferwise/pipelinewise/pull/293/files/a76d4e4609350f89b44a572c202c9d37b97dfe6b#r363235775

pipelinewise/fastsync/commons/tap_mysql.py
Samira-El: We can't trim after using hex because it would also trim zeros that are part of the data.

pipelinewise/fastsync/commons/tap_mysql.py
ivan-transferwise: `row = list(map(lambda val: codecs.encode(val.decode().rstrip('\x00').encode(), 'hex').decode("utf-8") if isinstance(val, bytes) else val, row))`
Didn't test it but should be smth like this

Samira-El: One liners are cool, but do they make an easily readable code?

PR #304, opened 2020-01-14
[AP-415] Create pipelinewise specific common temp directory

koszti:
## Description  

Currently every temp file created in the system `/tmp` directory but we'd like to have a pipelinewise specific temp directory in `~/.pipelinewise/tmp` to isolate pipelinewise temp files from system temp files. This PR makes it.

Temp files in scope:
* Fastsync `.csv.gz` files
* Generated runtime `config.json` files to run taps

## Extra
* The modified `pipelinewise/cli/schemas/target.json` JSON schema makes sure no target allowed to create with the name `tmp`. This check runs automatically by `validate` and `import` commands and from circeci as well.
* The generated target `config.json` automatically includes a new `temp_dir` property so every target can write their temp files into the same directory.

## Checklist

### DIFF
--- a/pipelinewise/cli/pipelinewise.py
+++ b/pipelinewise/cli/pipelinewise.py
@@ -263,7 +263,7 @@ class PipelineWise(object):
         else:
             self.config = {}
 
-    def get_temp_dir(self,):
+    def get_temp_dir(self):
         """
         Returns the tap specific temp directory
         """
--- a/pipelinewise/cli/pipelinewise.py
+++ b/pipelinewise/cli/pipelinewise.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 import os
 import shutil
-import tempfile
 import signal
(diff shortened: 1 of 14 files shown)


pipelinewise/cli/pipelinewise.py
```diff
@@ -256,6 +263,12 @@ def load_config(self):
         else:
             self.config = {}
 
+    def get_temp_dir(self,):
```
ivan-transferwise: Is that a valid syntax `get_temp_dir(self,)`?

koszti: seems valid but still typo, I will send new commit to fix it

koszti: thanks @ivan-transferwise fixed by 4de3d4aab8fcf3731520783f0492cfa96cd598bd

PR #309, opened 2020-01-17
[AP-421] Add stop_tap command

koszti:
## Description

This PR adds **stop_tap** command to PipelineWise. Usage: `pipelinewise stop_tap --target <TARGET_ID> --tap <TAP_ID>`

At the moment we need to stop long running taps manually by `kill -9 <PID>`.

This PR creates a pidfile for every running tap in `~/.pipelinewise/<target_id>/<tap_id>/pipelinewise.pid` so **stop_tap** can detect quickly the process id and can send SIGINT to the process that triggers a graceful exit.

## Extra

This will also allow us detecting if the tap is in running status much quicker as we do today. Currently we check the log history to find if the tap is running or not. In case of 150 taps the startup process takes at least 1 minute for each tap. Doing the same by checking if the pidfile exists is much quicker.

## Checklist

### DIFF
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import sys
+import pidfile
 from pathlib import Path
 
 from pkg_resources import get_distribution
@@ -18,6 +19,7 @@ venv_dir = os.path.join(pipelinewise_home, '.virtualenvs')
 commands = [
   'init',
   'run_tap',
+  'stop_tap',
   'discover_tap',
   'status',
   'test_tap_connection',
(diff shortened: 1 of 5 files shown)


pipelinewise/cli/__init__.py
```diff
@@ -63,7 +65,7 @@ def main():
             print("You must specify a project name using the argument --name")
             sys.exit(1)
 
-    if args.command == 'discover_tap' or args.command == 'test_tap_connection' or args.command == 'run_tap':
+    if args.command in ['discover_tap', 'test_tap_connection', 'run_tap', 'stop_tap']:
```
ivan-transferwise: Yas, nice change, now we are pythonic :D

PR #332, opened 2020-02-28
Allow SF reserved keywords as table names

ivan-transferwise:


Samira-El:
Hey, could you add a table that has reserved word in the e2e tests to test this change?

PR #357, opened 2020-03-16
Support for iam roles for snowflake fastsync

mashanm:
## Description

Support for AWS iam roles for snowflake fastsync.

Creating the boto3 client object without acces/secret keys will enable boto3 to look for all possible credentials . Creating the storage integration object in snowflake with necessary permission for the bucket will enable load data to snowflake to work without aws credentials

## Checklist

- [ ] Relevant documentation is updated including usag e instructions

### DIFF
--- a/pipelinewise/fastsync/commons/target_snowflake.py
+++ b/pipelinewise/fastsync/commons/target_snowflake.py
@@ -32,8 +32,10 @@ class FastSyncTargetSnowflake:
         self.connection_config = connection_config
         self.transformation_config = transformation_config
         self.s3 = boto3.client('s3',
-                               aws_access_key_id=self.connection_config['aws_access_key_id'],
-                               aws_secret_access_key=self.connection_config['aws_secret_access_key'])
+                               aws_access_key_id=self.connection_config.get('aws_access_key_id'),
+                               aws_secret_access_key=self.connection_config.get('aws_secret_access_key'),
+                               aws_session_token=self.connection_config.get('aws_session_token')
+                               )
 
     def open_connection(self):
         return snowflake.connector.connect(user=self.connection_config['user'],
@@ -139,37 +141,18 @@ class FastSyncTargetSnowflake:
         table_dict = utils.tablename_to_dict(table_name)
         target_table = table_dict.get('table_name') if not is_temporary else table_dict.get('temp_table_name')
         inserts = 0
-
-        aws_access_key_id = self.connection_config['aws_access_key_id']
-        aws_secret_access_key = self.connection_config['aws_secret_access_key']
         bucket = self.connection_config['s3_bucket']
 
-        master_key = self.connection_config.get('client_side_encryption_master_key', '')
-        if master_key != '':
-            sql = """COPY INTO {}."{}" FROM @{}/{}
+        sql = """COPY INTO {}."{}" FROM @{}/{}
                 FILE_FORMAT = (type='CSV' escape='\\x1e' escape_unenclosed_field='\\x1e' 
                 field_optionally_enclosed_by='\"' skip_header={} COMPRESSION='GZIP' BINARY_FORMAT='HEX') 
-            """.format(
-                target_schema,
-                target_table.upper(),
-                self.connection_config['stage'],
-                s3_key,
-                int(skip_csv_header),
-            )
-        else:
-            sql = """COPY INTO {}."{}" FROM 's3://{}/{}'
-                CREDENTIALS = (aws_key_id='{}' aws_secret_key='{}')
-                FILE_FORMAT = (type='CSV' escape='\\x1e' escape_unenclosed_field='\\x1e' 
-                field_optionally_enclosed_by='\"' skip_header={} COMPRESSION='GZIP' BINARY_FORMAT='HEX')
-            """.format(
-                target_schema,
-                target_table.upper(),
-                bucket,
-                s3_key,
-                aws_access_key_id,
-                aws_secret_access_key,
-                int(skip_csv_header),
-            )
+                """.format(
+                    target_schema,
+                    target_table.upper(),
+                    self.connection_config['stage'],
+                    s3_key,
+                    int(skip_csv_header),
+                )
 
         # Get number of inserted records - COPY does insert only
         results = self.query(sql)
--- a/pipelinewise/fastsync/mysql_to_snowflake.py
+++ b/pipelinewise/fastsync/mysql_to_snowflake.py
@@ -25,8 +25,6 @@ REQUIRED_CONFIG_KEYS = {
         'user',
         'password',
         'warehouse',
-        'aws_access_key_id',
-        'aws_secret_access_key',
         's3_bucket',
         'stage',
         'file_format'
--- a/pipelinewise/fastsync/postgres_to_snowflake.py
+++ b/pipelinewise/fastsync/postgres_to_snowflake.py
@@ -25,8 +25,6 @@ REQUIRED_CONFIG_KEYS = {
         'user',
         'password',
         'warehouse',
-        'aws_access_key_id',
-        'aws_secret_access_key',
         's3_bucket',
         'stage',
         'file_format'
--- a/pipelinewise/fastsync/s3_csv_to_snowflake.py
+++ b/pipelinewise/fastsync/s3_csv_to_snowflake.py
@@ -29,8 +29,6 @@ REQUIRED_CONFIG_KEYS = {
         'user',
         'password',
         'warehouse',
-        'aws_access_key_id',
-        'aws_secret_access_key',
         's3_bucket',
         'stage',
         'file_format'


koszti:
hey @mashanm , can you please resolve the conflicts? Should caused by some small changes that has been merged recently. thanks

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -154,21 +159,35 @@ def copy_to_table(self, s3_key, target_schema, table_name, is_temporary, skip_cs
...
+                    FILE_FORMAT = (type='CSV' escape='\\x1e' escape_unenclosed_field='\\x1e' 
+                    field_optionally_enclosed_by='\"' skip_header={} COMPRESSION='GZIP' BINARY_FORMAT='HEX')
+                    """.format(
+                    target_schema,
+                    target_table.upper(),
+                    bucket,
+                    s3_key,
+                    int(skip_csv_header),
+                )
+                self.query(sql)
```
koszti: hey @mashanm , can you please resolve the conflicts? I think it's here and caused by some small changes that has been merged recently. The`self.query(sql)` calls not required here. thanks

mashanm: Addressed the conflicts and moved the query call .

koszti:
I think these two lines are not required. Can you please remove them?
*  [fastsync/commons/target_snowflake.py:151](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L151)
*  [fastsync/commons/target_snowflake.py:152](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L152)

koszti:
There are some formatting issues reported by pylint. Can you please fix them?

You can reproduce the full log by running:
```
pip install pylint
pylint pipelinewise tests
```
Unfortunately the automated circle CI tests are currently not public and it's not reported automatically for external PRs

pipelinewise/fastsync/mysql_to_snowflake.py
```diff
@@ -25,8 +25,6 @@
         'user',
         'password',
         'warehouse',
-        'aws_access_key_id',
-        'aws_secret_access_key',
```
koszti: can you please make the same change in [postgres_to_snowflake](https://github.com/transferwise/pipelinewise/blob/master/pipelinewise/fastsync/postgres_to_snowflake.py) and [s3_csv_to_snowflake](https://github.com/transferwise/pipelinewise/blob/master/pipelinewise/fastsync/s3_csv_to_snowflake.py)

mashanm: Sure, Removed it in postgres and s3_csv as well

mashanm:
> I think these two lines are not required. Can you please remove them?
> 
> * [fastsync/commons/target_snowflake.py:151](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L151)
> * [fastsync/commons/target_snowflake.py:152](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L152)

mashanm:
> I think these two lines are not required. Can you please remove them?
> 
> * [fastsync/commons/target_snowflake.py:151](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L151)
> * [fastsync/commons/target_snowflake.py:152](https://github.com/transferwise/pipelinewise/blob/e8e4cdf2902b393d9a391e0a2f2bf37a955d40c5/pipelinewise/fastsync/commons/target_snowflake.py#L152)

 Removed those lines.

mashanm:
> There are some formatting issues reported by pylint. Some spaces are not at the right place, can you please fix them?
> 
> Unfortunately the automated circle CI tests are currently not public and it's not reported automatically for external PRs but you can reproduce the full log by running:
> 
> ```
> pip install pylint
> pylint pipelinewise tests
> ```

Formating issues were fixed

pipelinewise/fastsync/mysql_to_snowflake.py
koszti: thanks

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -31,9 +32,17 @@ class FastSyncTargetSnowflake:
...
+            self.use_iam_accesskeys = True
+        else:
+            self.use_iam_accesskeys = False
+
+        if self.use_iam_accesskeys:
+            self.s3 = boto3.client('s3',
+                                   aws_access_key_id=self.connection_config['aws_access_key_id'],
+                                   aws_secret_access_key=self.connection_config['aws_secret_access_key'])
+        else:
+            self.s3 = boto3.client('s3')
```
koszti: I've reviewed it again and compared it to pipelinewise-target-snowflake [here](https://github.com/transferwise/pipelinewise-target-snowflake/blob/a26fa575cd24bdf63603d845988db90fbc0c5229/target_snowflake/db_sync.py#L291) , where secrets, env vars, iam roles and aws_session_tokens are all supported. 

Can you please make the last change and modify it to be this code in sync with pipelinewise-target-snowflake? The code would be shorter and we'd add support for `aws_session_tokens`as well. I think it'd something like this:

```
        self.s3 = boto3.client(
            's3',
            aws_access_key_id=self.connection_config.get('aws_access_key_id'),
            aws_secret_access_key=self.connection_config.get('aws_secret_access_key'),
            aws_session_token=self.connection_config.get('aws_session_token')
        )
```

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -157,20 +164,33 @@ def copy_to_table(self, s3_key, target_schema, table_name, size_bytes, is_tempor
...
+                sql = """COPY INTO {}."{}" FROM 's3://{}/{}'
+                    FILE_FORMAT = (type='CSV' escape='\\x1e' escape_unenclosed_field='\\x1e' 
+                    field_optionally_enclosed_by='\"' skip_header={} COMPRESSION='GZIP' BINARY_FORMAT='HEX')
+                    """.format(
+                        target_schema,
+                        target_table.upper(),
+                        bucket,
+                        s3_key,
+                        int(skip_csv_header),
+                    )
```
koszti: I also noticed that this part can be significantly shorter and actually we can do everything in one COPY command similar if we follow the relevant part in pipelinewise-target-snowflake [here](https://github.com/transferwise/pipelinewise-target-snowflake/blob/a26fa575cd24bdf63603d845988db90fbc0c5229/target_snowflake/db_sync.py#L470) .

Creating a "name external stage object on s3" in snowflake is mandatory, so we can do everything by one SQL:

```
        sql = """COPY INTO {}."{}" FROM @{}/{}
            FILE_FORMAT = (type='CSV' escape='\\x1e' escape_unenclosed_field='\\x1e' 
            field_optionally_enclosed_by='\"' skip_header={} COMPRESSION='GZIP' BINARY_FORMAT='HEX') 
        """.format(
            target_schema,
            target_table.upper(),
            self.connection_config['stage'],
            s3_key,
            int(skip_csv_header),
        )
```
We don't need to pass credentials to s3 because it's ready to use in s3.

mashanm: In our setup, we were creating external stage without any credentials since fastsync always passed the keys while loading the data and credentials were not mandatory param while creating external object . Your current suggestion would affect this setup, Do you think it is still fine to go ahead to pass the external stage alone ?

pipelinewise/fastsync/commons/target_snowflake.py
mashanm: True it makes sense to initialise the s3 client without any conditions. The condition is also used while loading the data but if you think we could just pass the external stage assuming the credentials for reading from s3 is configured with the external stage, we can get rid of these conditions completely. I will wait for your comment (below) before removing it . Thanks for reviewing it .

koszti: > we could just pass the external stage assuming the credentials for reading from s3 is configured with the external stage
Yes, this is the correct behaviour. Please see more detailed answer below in the other comment.

pipelinewise/fastsync/commons/target_snowflake.py
koszti: You're right, your setup works with the current fastsync code but it's not really an expected behaviour. If you don't have an external stage to S3 with the correct credentials in snowflake then the first INCREMENTAL/LOG_BASED after fastsync will fail. INCREMENTAL and LOG_BASED are are using the [pipelinewise-target-snowflake connector](https://github.com/transferwise/pipelinewise-target-snowflake) and that requires the external stage. We should keep that behaviour here in fastsync as well.

So yes, please create an external stage for your s3 bucket and set the credentials in snowflake or use SF storage integration. Here we should use s3 directly only to upload the csv files.

mashanm: Thanks for the explanation, Adapted the fastsync as per your recommendation. Will try to update the documentation in the coming days :)

koszti: hey @mashanm the related documentation updated by 5548489c73cf2998396c0f2e9e0d4e273a2cc7b4

koszti:
merged. thank you for your work on this @mashanm

PR #359, opened 2020-03-18
update install.sh

Samira-El:
## Description

Fix dependency conflicts
Add conflict checks in circleci and exit with 1 if there are any. 
stop mounting .virtualenvs folder

## Checklist


dev-project/docker-compose.yml
```diff
@@ -29,11 +29,8 @@ services:
       DB_TARGET_POSTGRES_USER: ${DB_TARGET_POSTGRES_USER}
       DB_TARGET_POSTGRES_PASSWORD: ${DB_TARGET_POSTGRES_PASSWORD}
       DB_TARGET_POSTGRES_DB: ${DB_TARGET_POSTGRES_DB}
-
     volumes:
       - ../:/opt/pipelinewise
-      # don't create container specific .virtualenvs files on host
-      - /opt/pipelinewise/dev-project/.virtualenvs/
```
koszti: removing this will basically create the .virtualenvs on the host and will overwrite the content on the host. Is this what we want?

`- ../:/opt/pipelinewise` <- this mounts the entire pipelinewise repository
- /opt/pipelinewise/dev-project/.virtualenvs/ <- this is an exception rule to do NOT mount a specific sub-dir from host, but create an empty dir in the container instead

The idea is detailed at https://stackoverflow.com/a/37898591

Samira-El: Ah good catch! I misunderstood that line thinking it actually mounts the folder from host to the container. 
Thanks for the explanation, I'll put it back.

Samira-El: Fixed in [commit](https://github.com/transferwise/pipelinewise/pull/359/commits/9adff2219e2b0e9b83b44f328878ed81f15bfaad).

PR #368, opened 2020-03-23
add_metadata_columns is not defined in inheritable_config

ReptilianBrain:
add_metadata_columns doesn't seem to be defined by default when building the tap inheritable_config.json file. Making it harder to configure it later inside a tap .yml configuration.

This PR add the value inside the tap_inheritable_config dictionary.

## Description

<< Write here the description of your PR. Without description won't be approved! >>

## Checklist


koszti:
@ReptilianBrain can you please add this to the relevant test case as well at https://github.com/transferwise/pipelinewise/blob/940b2ec9c43ae0339133696994d943934bdc197b/tests/units/cli/test_config.py#L273 ?

Adding `'validate_records': False` as the expected value after the line above should do the trick.

PR #339, opened 2020-03-05
[AP-549] Support reserved words as table names

Samira-El:
## Description

Supporting reserved words as table names when syncing from PG to SF, this includes PG reserved words used as table names in PG DB and SF reserved words used as table names.

### Example

Source table:
```
DROP TABLE IF EXISTS "ORDER" CASCADE;

CREATE TABLE "ORDER"
(
    id serial primary key,
    cvarchar varchar,
    created_at timestamp default current_timestamp
);

insert into "ORDER" (cvarchar) values ('A'), ('B'), ('C'), ('D'), ('E'), ('F'), ('G');

```

Replication config

```
      - table_name: "order"
        replication_method: "INCREMENTAL"
        replication_key: "created_at"
        transformations:
          - column: 'cvarchar'
            type: 'MASK-HIDDEN'
```

Table in SF
image

## Checklist

- [X] Description above provides context of the change
- [x] Unit tests coverage for changes (not needed for documentation changes)
- [x] Bumping version in `setup.py` is an individual PR and not mixed with feature or bugfix PRs
- [x] Commit message/PR title starts with `[AP-NNNN]` if applicable. AP-NNNN = JIRA ID
- [x] Branch name starts with `AP-NNN` if applicable. AP-NNN = JIRA ID
- [ ] Commits follow "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"
- [ ] Relevant documentation is updated including usage instructions

PR #372, opened 2020-03-27
Added double quotes to prevent name clash between table and keywords in Redshift

cjrberg:
## Description

I experienced some issues running fast sync on MySQL to Redshift, having a table called user gave some problems due to name clash in Redshift. To fix the problem I added double quotes around the table name to avoid a name clash.

## Checklist

### DIFF
--- a/pipelinewise/fastsync/commons/target_redshift.py
+++ b/pipelinewise/fastsync/commons/target_redshift.py
@@ -190,4 +190,4 @@ class FastSyncTargetRedshift:
 
         # Swap tables and drop the temp tamp
         self.query('DROP TABLE IF EXISTS {}.{}'.format(schema, target_table))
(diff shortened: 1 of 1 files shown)


koszti:
hi @cjrberg thanks for sending this. Unfortunately this happens at a few more other places as well and we should fix all occasions together.

Recently we've made the same thing for Snowflake target and we need to do it for Redshift. Here is the full list what changed in snowflake to handle reserved words correctly:
* FastSync to Snowflake: https://github.com/transferwise/pipelinewise/pull/339
* Target-Snowflake: https://github.com/transferwise/pipelinewise-target-snowflake/pull/65

Some good news that the redshift and snowflake targets are very similar and making it for redshift is relatively straightforward.

koszti:
Using reserved words as table and column names has now been fixed across every component.
It's available in master and soon in the next PPW 0.16.0 release.

PR #397, opened 2020-05-05
[AP-655] Fixed issue when replication slot not created correctly

koszti:
## Description

FastSync tap-postgres sometimes not creating replication slot names correctly and sometimes generating invalid SQL. This PR fixing it.

## Checklist


setup.py
```diff
@@ -25,7 +25,7 @@
           'attrs==17.4.0',
           'idna==2.7',
           'PyMySQL==0.7.11',
-          'psycopg2==2.8.2',
+          'psycopg2-binary==2.8.5',
```
Samira-El, line 28: Why the switch to the binary psycopg2?

koszti: It doesn't require the extra C compiler and other OS packages on the host and it's reducing the size of the docker images and built times. tap-postgres already switched to `psycopg2-binary` and this keeps the two components in sync.

PR #414, opened 2020-05-14
Ap 372

koszti:
## Description

This PR is a proposal.
It reads an optional `session_sqls` parameter from the connection config and running each of them automatically right after the mysql/mariadb connection has made.

Using this approach we don't hardcode session parameters, every tap has different values and we support different different session parameter names used by the different version of mariadb/mysql.

```
id: "fx_ap372"
name: "AP-372 fx test"
type: "tap-mysql"
owner: "louis.pieterse@transferwise.com"
sync_period: "*/5 * * * *"
db_conn:
  host: "localhost"
  port: 3306
  user: my_user
  password: secret
  dbname: "fx"
  session_sqls:
    - SET SESSION max_statement_time=0
    - SET SESSION wait_timeout=28800
    - ... some more SQL ...
``` 

If we find that this option is the right one then the same approach can be added to tap-mysql and will be covered by tests.

## Checklist

### DIFF
--- a/dev-project/pipelinewise-config/tap_postgres_logical.yml
+++ b/dev-project/pipelinewise-config/tap_postgres_logical.yml
@@ -14,6 +14,10 @@ owner: "somebody@foo.com"              # Data owner to contact
 # ------------------------------------------------------------------------------
 db_conn:
   host: "db_postgres_source"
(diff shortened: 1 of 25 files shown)


koszti:
duplicate of #415

PR #436, opened 2020-06-04
[AP-689] Disable incremental for tap_mongodb

Samira-El:
## Description

Disable the incremental method with the help of the validation against json schema.
Given that the we can only use Draft4 valid schema currently, I went with the implication pattern to put a condition on the replication methods that can be used for tap_mongodb. 

PS: Example of [the implication pattern](https://stackoverflow.com/questions/51203606/jsonschema-attribute-conditionally-required-depends-on-parent-object?noredirect=1&lq=1).

## Checklist

### DIFF
--- a/pipelinewise/cli/schemas/tap.json
+++ b/pipelinewise/cli/schemas/tap.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "http://json-schema.org/draft-04/schema#",
   "definitions": {
     "schema": {
       "type": "object",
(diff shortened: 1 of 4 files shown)


koszti:
would you still prefer to upgrade draft7 everywhere and start using if-then conditions later?

Samira-El:
> do you still prefer to upgrade draft7 everywhere and start using if-then conditions?
> 
> would you do that now or it’s fine to merge it with the implication patter and we will refactor later if required and if draft7 is not backward compatible?

Thanks, I'll keep the current implementation for now and make the necessary changes later when we upgrade to draft7.

PR #444, opened 2020-06-11
[AP-726] Fix bad dates FS Mysql -> SF

Samira-El:
## Problem

When running FS from Mysql/MariaDB to snowflake, some bad dates in the source DB are not accepted by Snowflake, such dates are: 
- 31st day of  February: `2017-02-31 00:00:00` 
- day/month 0: `2017-00-00 00:00:00`
- year 0: `0000-03-26 00:36:37`
- year > 9999: `52347-09-19 17:18:40`

The current implementation queries datetime types with the following query:
```
select ..., REPLACE(nullif(`date_col`, "0000-00-00 00:00:00"), '\n', ' ') date_col, ... from table;
```
## Proposed changes

When querying a date(time) column without changing its type, the server returns valid date even if the inserted date is invalid, for example:

2 rows were inserted where date_col is `2017-02-31 00:00:00` and  `2000-00-00 17:18:40`.

Running `select concat(date_col, ''), date_col from table` will return:

```
concat(date_col, '') |   date_col
------------------------------------------------------------
2017-02-31 00:00:00  |  2017-03-03 00:00:00
2000-00-00 17:18:40  |  1999-11-30 17:18:40 
```

Using `replace` in the current query makes the server return the invalid date, dropping the `replace` when date type columns are processed avoids the server returning invalid dates.

Here is an example of what the following query against the source returns:

```
select
        CONCAT(date_created, ''),
       date_created as date_created_as_date,
       REPLACE(nullif(`date_created`, "0000-00-00 00:00:00"), '\n', ' ') current_impl,
       nullif(`date_created`, "0000-00-00 00:00:00") current_impl_without_replace
from weight_unit;
```

image

Use replace only for "text" type columns.

I have added new rows to sql test data for the E2E tests.

## Types of changes

## Checklist


pipelinewise/fastsync/commons/tap_mysql.py
```diff
@@ -200,24 +200,24 @@ def get_table_columns(self, table_name):
...
                             WHEN data_type IN ('blob', 'tinyblob', 'mediumblob', 'longblob', 'geometry')
-                                    THEN concat('hex(`', column_name, '`)')
+                                    THEN CONCAT('REPLACE(hex(`', column_name, '`)', ", '\n', ' ')")
                             WHEN data_type IN ('binary', 'varbinary')
-                                    THEN concat('hex(trim(trailing CHAR(0x00) from `',COLUMN_NAME,'`))')
+                                    THEN concat('REPLACE(hex(trim(trailing CHAR(0x00) from `',COLUMN_NAME,'`))', ", '\n', ' ')")
                             WHEN data_type IN ('bit')
                                     THEN concat('cast(`', column_name, '` AS unsigned)')
                             WHEN data_type IN ('datetime', 'timestamp', 'date')
                                     THEN concat('nullif(`', column_name, '`,"0000-00-00 00:00:00")')
```
koszti, line 215: This is somehow related to another recently reported issue of mysql 8 incompatibility.

What do you think can we use `str_to_date` as suggested in the link above and would that still work with mariadb/mysql5.x? And if yes, what do you think, can we add `str_to_date` into this PR or it should be addressed separately?

It'd be something like:
`nullif(`', column_name, '`, str_to_date("0000-00-00 00:00:00", "%Y-%m-%d %H:%i:%s"))


what do you think can we use

Samira-El: Seems fine to me, killing two birds with one stone.
I ran few queries and this doesn't impact change the result.

Samira-El: Will make a commit.

Samira-El: I used the format `%Y-%m-%d %T`, `%T` is the equivalent of `%H:%i:%S`.

PR #467, opened 2020-07-13
[AP-531] Add generic alert sender and slack alert handler

koszti:
## Problem

PPW doesn't have alerting mechanism and integration with alerting tools.

## Proposed changes

Introducing a new top level PPW [config.yml](https://github.com/transferwise/pipelinewise/blob/2cf15fbeb3dced857b04932fa1d2c1f7d7a758d2/pipelinewise/cli/samples/config.yml) with a new `alert_handlers` key:
```
---
alert_handlers:
  slack:
    token: "slack-token"
    channel: "#slack-channel"
```

If `run_tap` or `sync_tables` PPW commands fail, PPW checks if there is any configured `alert_handler` in [config.yml](https://github.com/transferwise/pipelinewise/pull/467/files#diff-27c4581f4324a9d1d6bc358c11a0b6d4). If it finds any then it calls the common `send` method of every handler to send the alert to the right destination(s).

### How does it work?

Every **alert handler** is an implementation of [BaseAlertHandler](https://github.com/transferwise/pipelinewise/blob/2cf15fbeb3dced857b04932fa1d2c1f7d7a758d2/pipelinewise/cli/alert_handlers/base_alert_handler.py) class, that allows to develop new handlers and new alerting tool integrations. This PR implements one alert handler, the [SlackAlertHandler](https://github.com/transferwise/pipelinewise/blob/2cf15fbeb3dced857b04932fa1d2c1f7d7a758d2/pipelinewise/cli/alert_handlers/slack_alert_handler.py)

**Routing** from the above `config.yml` to the actual alert handler classes is implemented by [alert_sender.py](https://github.com/transferwise/pipelinewise/blob/2cf15fbeb3dced857b04932fa1d2c1f7d7a758d2/pipelinewise/cli/alert_sender.py)

### To develop new alert handlers:
1. Define the configuration schema for the new alert handler in [config.json](https://github.com/transferwise/pipelinewise/blob/0f596c9abb50af60e3d8f066934649938949c5ef/pipelinewise/cli/schemas/config.json#L25).
2. Create a new alert handler class that implements the [BaseAlertHandler](https://github.com/transferwise/pipelinewise/pull/467/files#diff-cd3b30370a403a0199ae39d978b23e06) class. The `config` dictionary passed automatically from the `config.yml`
3. Register the new alert handler by adding it the list in [alert_sender.py](https://github.com/transferwise/pipelinewise/blob/2cf15fbeb3dced857b04932fa1d2c1f7d7a758d2/pipelinewise/cli/alert_sender.py#L23)

### To disable alerts on certain taps:

Sometimes you might want to disable the alerts on certain taps even if the tap failed. In this case add `send_alert: False` to the tap config YAML. 

## Types of changes

Documentation will be added once we're OK with the proposal. 

## Checklist

Documentation will be added once we're OK with the proposal.


pipelinewise/cli/alert_handlers/base_alert_handler.py
```diff
@@ -0,0 +1,28 @@
+"""
+PipelineWise CLI - Base class of alert handlers
+"""
+
+
+# pylint: disable=too-few-public-methods
+class BaseAlertHandler:
```
Samira-El, line 7: I recommend configuring this class to be an abstract class with the help `from abc import ABC` and then decorating `send` with `@abstractmethod`.
That way nobody can create direct instances of `BaseAlertHandler` and any class that  inherits from `BaseAlertHandler` would be forced to implement `send` and avoid getting `NotImplementedError` at runtime.

Samira-El: Check the doc https://docs.python.org/3/library/abc.html

koszti: thanks, addressed by https://github.com/transferwise/pipelinewise/pull/467/commits/d4cd6e022a4afa7414d5f4055c571520458905db

pipelinewise/cli/alert_handlers/errors.py
```diff
@@ -0,0 +1,27 @@
...
+PipelineWise CLI - Alert handler exceptions
+"""
+
+
+class NotImplementedAlertHandlerException(Exception):
+    """
+    Exception to raise when attempted to use a not implemented alert handler class
+    """
+    def __init__(self, *args, **kwargs):
+        Exception.__init__(self, *args, **kwargs)
```
Samira-El, line 11: Please use `super().__init__` instead of `Exception.__init__`, there are the same however, the practice is to use `super` to call the parent class.

pipelinewise/cli/alert_sender.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Get an alert handler from the alert handlers dict
+
+        Args:
+            alert_handler_type: type of the alert handler (slack, rollbar, etc.)
+
+        Returns:
+            AlertHandler tuple
+        """
+        if alert_handler_type in self.alert_handlers.keys():
```
Samira-El, line 78: No need to call `.keys()`,  `alert_handlers` ia dictionary so `if alert_handler_type in self.alert_handlers:` is enough.

pipelinewise/cli/alert_sender.py
```diff
@@ -0,0 +1,129 @@
...
+
+        Args:
+            message: alert text message
+            level: alert level
+            exc: optional exception that triggered the alert
+
+        Returns:
+            Dictionary with number of successfully sent alerts
+        """
+        sents = [self.send_to_handler(handler_type, message, level, exc) for handler_type in self.alert_handlers.keys()]
```
Samira-El, line 128: Same comment about `keys()`

pipelinewise/cli/config.py
```diff
@@ -54,7 +63,7 @@ def from_yamls(cls, config_dir, yaml_dir='.', vault_secret=None):
             target_id = target_data['id']
 
             # Check if a target with same ID already exists
-            # exit with error, otherwise, we would override the previous one and that would go unnoticed.
+            # exit with error, oterwise, we would override the previous one and that would go unnoticed.
```
Samira-El, line 66: `otherwise` is the correct one :lel:

koszti: fixed by https://github.com/transferwise/pipelinewise/pull/467/commits/0ebaab95c33cce5ac03d60ccd9e90dad7c6c9e82

pipelinewise/cli/alert_sender.py
koszti: fixed by https://github.com/transferwise/pipelinewise/pull/467/commits/68ef940569dda204cc08e12daa731365799484bb

pipelinewise/cli/alert_sender.py
koszti: fixed by https://github.com/transferwise/pipelinewise/pull/467/commits/d704908b6a2a80ee98b572d8ac47e8f94f0555af

pipelinewise/cli/alert_handlers/errors.py
koszti: fixed by https://github.com/transferwise/pipelinewise/pull/467/commits/f74caa958765671301a56f81cfc7643bd269d81a

PR #494, opened 2020-07-30
Override extra_config_keys with user config in tap_config

NiallRees:
The config was overriding the user's own config, and already exists here: https://github.com/transferwise/pipelinewise-tap-zendesk/blob/master/tap_zendesk/default_config.json

### DIFF
--- a/pipelinewise/cli/tap_properties.py
+++ b/pipelinewise/cli/tap_properties.py
@@ -151,11 +151,7 @@ def get_tap_properties(tap=None, temp_dir=None):
             'default_data_flattening_max_level': 0
         },
         'tap-zendesk': {
-            'tap_config_extras': {
-                'rate_limit': 1000,
-                'max_workers': 10,
-                'batch_size': 50
-            },
+            'tap_config_extras': {},
             'tap_stream_id_pattern': '{{table_name}}',
             'tap_stream_name_pattern': '{{table_name}}',
             'tap_catalog_argument': '--catalog',


koszti:
Instead of switching the dict unpacking logic, can you please just remove the extra args from the [tap_properties.py](https://github.com/transferwise/pipelinewise/blob/master/pipelinewise/cli/tap_properties.py#L155-L157) ? That should do the same thing. If internal `extra_args` not provided to the tap, then it should use the tap's [default_config.json](https://github.com/transferwise/pipelinewise-tap-zendesk/blob/master/tap_zendesk/default_config.json) 

Switching the dict unpacking logic probably makes sense as well, but that'd directly impact every tap-target combination and that kind of change requires a very extensive testing. Would be cool to make this PR tap-zendesk specific only.

NiallRees:
> Instead of switching the dict unpacking logic, can you please just remove the extra args from the [tap_properties.py](https://github.com/transferwise/pipelinewise/blob/master/pipelinewise/cli/tap_properties.py#L155-L157) ? That should do the same thing. If internal `extra_args` not provided to the tap, then it should use the tap's [default_config.json](https://github.com/transferwise/pipelinewise-tap-zendesk/blob/master/tap_zendesk/default_config.json)
> 
> Switching the dict unpacking logic probably makes sense as well, but that'd directly impact every tap-target combination and that kind of change requires a very extensive testing. Would be cool to make this PR tap-zendesk specific only.

Sure

koszti:
thanks! 🙇

PR #514, opened 2020-08-26
[AP-774] Properly support Mysql Time column

Samira-El:
## Problem

Currently, pipelinewise FastSync and singer replication are not aligned on how to handle `Time` column in Mysql:
* FastSync lacks mapping so Time is by default mapped to Varchar
* Singer replication on the other hand interprets `time` as a datetime.

## Proposed changes

* FastSync: map `time` to the right equivalent in targets.
* Singer replication: use https://github.com/transferwise/pipelinewise/pull/514 

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc

### DIFF
--- a/tests/end_to_end/helpers/db.py
+++ b/tests/end_to_end/helpers/db.py
@@ -33,6 +33,7 @@ def run_query_mysql(query, host, port, user, password, database):
                          user=user,
                          password=password,
                          database=database,
+                         charset="utf8mb4",
                          cursorclass=pymysql.cursors.Cursor) as cur:
         cur.execute(query)
         if cur.rowcount > 0:
--- a/tests/end_to_end/helpers/db.py
+++ b/tests/end_to_end/helpers/db.py
@@ -33,7 +33,7 @@ def run_query_mysql(query, host, port, user, password, database):
                          user=user,
                          password=password,
                          database=database,
-                         charset="utf8mb4",
+                         charset='utf8mb4',
                          cursorclass=pymysql.cursors.Cursor) as cur:
         cur.execute(query)
         if cur.rowcount > 0:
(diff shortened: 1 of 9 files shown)


tests/end_to_end/helpers/db.py
```diff
@@ -33,6 +33,7 @@ def run_query_mysql(query, host, port, user, password, database):
                          user=user,
                          password=password,
                          database=database,
+                         charset='utf8mb4',
```
koszti, line 36: just wondering, is there any specific reason of adding this?

Samira-El: Sure, if you look at the changes I've made to `test_target_postgres/snowflake.py`, you can see that the query entails inserting some special characters, if I don't specify a utf8 charset, there is an encoding error.

Samira-El: yu can see inhttps://github.com/transferwise/pipelinewise/blob/fa24440ea3cbdf0baff57a80b764cfe92213010e/tests/db/tap_mysql_data.sql#L32 that the table's DDL sets charset to utf8.

Samira-El: utf8mb4 is the better cousin of utf8 ;)

Samira-El: Let me know if I should do it differently.

koszti: I was only wondering why it didn't fail earlier without specifying charset in the `run_query_mysql`. But the default charset on the table level is probably the answer as you said.

Thanks you

koszti (approved):

PR #521, opened 2020-09-03
[AP-369] Handle timestamps out of range in PG source

Samira-El:
## Problem

PG Source can contain dates outside of the range [`0001-01-01 00:00:00.000000`, `9999-12-31 23:59:59.999999`].
These dates cause FS ans Singer replication to fail or results in invalid data.

### FS
Use push down processing to check the dates range and fallback to `9999-12-31 23:59:59.999` if they are outside of the above range.

### Singer replication
Use https://github.com/transferwise/pipelinewise-tap-postgres/pull/62  

*Note*: 
Having the microseconds in the fallback date `9999-12-31 23:59:59.999` be only 999 instead of 999999 is a measure to avoid an issue with target snowflake, see [here](https://github.com/snowflakedb/snowflake-connector-python/issues/392)

In Snowflake, fetching the value `9999-12-31 23:59:59.999` from a timestamp column returns this datetime instance `datetime.datetime(9999, 12, 31, 23, 59, 59, 998993)` 

## Types of changes

## Checklist


pipelinewise/fastsync/commons/tap_postgres.py
```diff
@@ -274,6 +274,10 @@ def get_table_columns(self, table_name):
                     WHEN data_type = 'ARRAY' THEN 'array_to_json("' || column_name || '") AS ' || column_name
                     WHEN udt_name = 'time' THEN 'replace("' || column_name || E'"::varchar,\\\'24:00:00\\\',\\\'00:00:00\\\') AS ' || column_name
                     WHEN udt_name = 'timetz' THEN 'replace(("' || column_name || E'" at time zone \'\'UTC\'\')::time::varchar,\\\'24:00:00\\\',\\\'00:00:00\\\') AS ' || column_name
+                    WHEN udt_name in ('timestamp', 'timestamptz') THEN
+                       'CASE WHEN "' ||column_name|| E'" < \\'0001-01-01 00:00:00.000000\\' '
+                            'OR "' ||column_name|| E'" > \\'9999-12-31 23:59:59.999\\' THEN \\'9999-12-31 23:59:59.999\\' '
```
koszti, line 279: To avoid some edge cases dates need be in milliseconds otherwise dates like `0000-01-01 00:00:00.000123` not converted to the fallback date in milliseconds.

koszti (changes requested):

pipelinewise/fastsync/commons/tap_postgres.py
Samira-El: Thanks for the feedback!
How did you manage to insert `0000-01-01 00:00:00.000123` into the pg table tho? cuz I'm getting `date/time field value out of range:` when running `insert`

koszti: I was only reading the code but didn't try it. I spotted the inconsistency of milliseconds vs microseconds and the potential edge case if the value in the db is `0000-01-01 00:00:00.000123`.

Maybe postgres doesn't allow it, but can we still use milliseconds to be mathematically correct? what do you think?

Samira-El: update in https://github.com/transferwise/pipelinewise/pull/521/commits/d04465864347890749944f1983061d19435eacf8

koszti (approved):

singer-connectors/tap-postgres/requirements.txt
```diff
@@ -1 +1 @@
-pipelinewise-tap-postgres==1.6.3
+-e git+https://github.com/transferwise/pipelinewise-tap-postgres.git@AP-369#egg=pipelinewise-tap-postgres
```
koszti, line 1: can you please bump ppw-tap-postgres to 1.6.4? It's now available on PyPI

koszti (changes requested):

singer-connectors/tap-postgres/requirements.txt
Samira-El: Sure

koszti (approved):

lucaswiley:
@Samira-El Does this resolve https://github.com/transferwise/pipelinewise/issues/504 ? Thanks

Samira-El:
Hey @lucaswiley, yes, it does.

lucaswiley:
@Samira-El Thanks for beating me to it.

Issue #529, opened 2020-09-15
feature: generate valid PostgreSQL replication slot name when db name has invalid chars


hgrsd:
When using logical replication from a PostgreSQL tap, the name of the replication slot used to read the WAL is automatically generated based on the database name and tap name. See [this piece of code](https://github.com/transferwise/pipelinewise-tap-postgres/blob/c0eb14ebfe14b10ad208d10c5d1d7f1bf5b8be25/tap_postgres/sync_strategies/logical_replication.py#L402). 

We have run into issues because some of our source databases had names that include a hyphen `-`, which is not allowed in the replication slot name.

More generally, the issue is that the set of characters allowed in Postgres database does not match the set of characters allowed in replication slot names (i.e. `[a-z0-9_]`).

Possible solutions:
- replace invalid characters either by an underscore or their unicode code point (see an example of this [here](https://github.com/zalando/patroni/pull/277))
- allow user to specify replication slot name as part of the `tap.yaml` configuration

The first option is probably simplest to implement. I would be happy to work on this and open a PR.

Looking forward to hearing what you think. Thanks in advance!

koszti:
hey, thanks for this. Option one sounds like the best option: ubderscores, lowercase and trunc to max 64 chars similar to the patroni example.

Can you please send a PR? I’m sure we can merge that quickly.

Can you please make sure to add a few new test cases as well to cover ? You can find examples at https://github.com/transferwise/pipelinewise/blob/6be4ee8e4ea8096deb955f823cf6ca81476135e1/tests/units/fastsync/commons/test_fastsync_tap_postgres.py#L35

koszti:
One more thing: we need to make sure that ppw-tap-postgres behaves the same here at https://github.com/transferwise/pipelinewise-tap-postgres/blob/c0eb14ebfe14b10ad208d10c5d1d7f1bf5b8be25/tap_postgres/sync_strategies/logical_replication.py#L402

Can you please send PR to tap-pg as well with the same change?

hgrsd:
Thank you @koszti for your reply. I've opened two PRs as per your suggestion; let me know if you want any more work to be done on them.

sroze:
Out of interest @koszti, do you have any plan to prevent/reduce this duplicated code?

koszti:
fully fixed and released as part of 0.23.0 

[...]

PR #530, opened 2020-09-16
[FIX] - Ensure generated Postgres replication slot name is valid

hgrsd:
## Problem

When using logical replication from a Postgres tap, Pipelinewise generates the name of the Postgres replication slot used from a combination of prefix, database name, and tap name. This does not work if any of the constituent parts contain characters that are not valid in Postgres replication slot names, or if the full generated name is more than 64 characters long.

## Proposed changes

Replace invalid characters with an underscore and truncate the generated name at 64 characters. See https://github.com/transferwise/pipelinewise/issues/529. A similar PR is open on the tap-postgres repository: https://github.com/transferwise/pipelinewise-tap-postgres/pull/63.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [X] Description above provides context of the change
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Unit tests for changes (not needed for documentation changes)
- [X] Commits follow "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -2,6 +2,7 @@ import datetime
 import decimal
 import gzip
 import logging
+import re
 import psycopg2
 import psycopg2.extras
 
@@ -40,7 +41,11 @@ class FastSyncTapPostgres:
         # Convert None to empty string
         else:
             tap_id = ''
-        return f'{prefix}_{dbname}{tap_id}'.lower()
+
+        slot_name = f'{prefix}_{dbname}{tap_id}'.lower()
+
+        # Replace invalid characters and truncate to a maximum of 64 characters; required by Postgres
+        return re.sub('[^a-z0-9_]', '_', slot_name)[:64]
 
     def open_connection(self):
         """
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -44,8 +44,8 @@ class FastSyncTapPostgres:
 
         slot_name = f'{prefix}_{dbname}{tap_id}'.lower()
 
-        # Replace invalid characters and truncate to a maximum of 64 characters; required by Postgres
-        return re.sub('[^a-z0-9_]', '_', slot_name)[:64]
+        # Replace invalid characters to ensure replication slot name is in accordance with Postgres spec
+        return re.sub('[^a-z0-9_]', '_', slot_name)
 
     def open_connection(self):
         """
--- a/singer-connectors/tap-postgres/requirements.txt
+++ b/singer-connectors/tap-postgres/requirements.txt
@@ -1,1 +1,1 @@
-pipelinewise-tap-postgres==1.6.3
+pipelinewise-tap-postgres==1.6.4
--- a/tests/units/fastsync/commons/test_fastsync_tap_postgres.py
+++ b/tests/units/fastsync/commons/test_fastsync_tap_postgres.py
@@ -6,6 +6,7 @@ class FastSyncTapPostgresMock(FastSyncTapPostgres):
     """
     Mocked FastSyncTapPostgres class
     """
+
     def __init__(self, connection_config, transformation_config=None):
         super().__init__(connection_config, transformation_config)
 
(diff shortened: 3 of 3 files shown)


koszti (changes requested):
...

hgrsd:
Hmm, good questions. Thinking this through (without a lot of familiarity with Pipelinewise!), my view is probably as follows.

- I don't think we can call external functions from a JSON schema
- Adding a maximum length to fields in the JSON schema comes with its own problems. Three concerns come to mind: 
  - As you say, if we decide a limit on the database name length we might be excluding databases with long names even though we'd be able to generate a replication slot name by truncating the name.
  - We might be invalidating existing replication slots that people are currently using because either their database or tap id might be longer than the limit we decide, even though in _combination_  (and in combination with the prefix) the generated name would not amount to more than 64 characters.
   - We'd be applying a limit that is specific to one feature of Postgres to users who do not use the logical replication method at all

Perhaps the "right" solution is in fact to be less smart:
- we replace invalid characters in the database name by underscores (because database names can be awkward to change, especially in production environments)
- we do _not_ replace invalid characters in the tap id: we can add the valid characters to the Postgres tap documentation; and if the user makes a mistake the tap id is easy to change in the yaml file with zero repercussions
- we do _not_ truncate to 64 characters: again, we can add this limit to the documentation, and if the user has a combination of dbname & tap id that is > 64 characters, they will have to decide how to deal with it, most likely by shortening their chosen tap id

This limits the interference with the user's data to a minimum, ensures that we don't need any database renames (which to me would be the key benifit of this PR), is backwards compatible (because we know that there can be no invalid replication slots that are currently being used), and means that any issue can be resolved by the user in their yaml configuration. 

What do you think?

koszti:
Related PR for ppw-tap-postgres: https://github.com/transferwise/pipelinewise-tap-postgres/pull/63. We'll need to release a new version to PyPI and bump the postgres tap [here](https://github.com/transferwise/pipelinewise/blob/master/singer-connectors/tap-postgres/requirements.txt).

hgrsd:
Thanks @koszti. How do you and @Samira-El feel about the approach outlined above? We could make it slightly smarter by truncating _only the database name_ if the total length of the generated replication slot name is more than 64 characters. That way the user will retain the ability to prevent name collisions because we make sure not to interfere with their chosen tap id.

koszti:
There is no perfect solution.

I'd replace the invalid characters both in dbname and tap_id to keep the behaviours in sync.

And I wouldn't truncate the generated slot name to 64 characters. If it's too long then it will fail at runtime. We can't check the generated length by jsonschemas at import time but I think that's still fine. Having too long replication slot name is a very rare situation and we wouldn't break anything. If it's too then we need to shorten the tap_id.

@Samira-El what do you think?

Samira-El:
Hey all,

Thanks btw for the PR!

According to the section **_26.2.6.1. Querying And Manipulating Replication Slots_** in the [PG doc](https://www.postgresql.org/docs/current/warm-standby.html#STREAMING-REPLICATION-SLOTS), the only restriction on the slot name is it should only contain lower-case letters, numbers, and the underscore character.

So having mentioned that, I agree with Peter, I'd replace the invalid chars in the slot name.
The PG doc doesn't mention any restriction on the slot length, or maybe I've missed it somewhere, so I wouldn't enforce the 64 chars limit.

hgrsd:
Thank you both. I've updated this PR and the PR in the postgres tap repo so that we no longer truncate the name to 64 chars. Thanks for helping figure out the best approach to this!

koszti:
I merged https://github.com/transferwise/pipelinewise-tap-postgres/pull/63 but there are a couple of other pending PRs in tap-postgres: https://github.com/transferwise/pipelinewise-tap-postgres/pull/62 and https://github.com/transferwise/pipelinewise-tap-postgres/pull/64

We need to decide if we want to put all of them into next tap-postgres release or it's better to create a new one only for this PR.

hgrsd:
> We need to decide if we want to put all of them into next tap-postgres release or it's better to release now onto PyPI so we have a new version only for this PR.

I'm unfamiliar with the release strategy for this project so not sure if I have a very useful point of view, apart from the fact that for us it'd be great if this could be released relatively soon. It'll save us from having to rename a database 😁 .

Do ping me if you need any action from me on this PR.

koszti:
@hgrsd tap-postgres 1.6.4 is now available on PyPI. Can you please bump the connecter [here](https://github.com/transferwise/pipelinewise/blob/master/singer-connectors/tap-postgres/requirements.txt) ?

koszti (approved):

PR #579, opened 2020-11-27
Add ssl support to mongodump.

novotl:
## Problem

- The ssl option is not taken into account when using mongodump which results in hanging connection.

## Proposed changes

Take the `ssl` value from connection config into account when using `mongodump`. The same value is already used here https://github.com/transferwise/pipelinewise/blob/5b8e9fa80f620179a21777ea427114f0ace4be48/pipelinewise/fastsync/commons/tap_mongodb.py#L123

## Types of changes

## Checklist

### DIFF
--- a/pipelinewise/fastsync/commons/tap_mongodb.py
+++ b/pipelinewise/fastsync/commons/tap_mongodb.py
@@ -298,6 +298,9 @@ class FastSyncTapMongoDB:
         if self.connection_config.get('replica_set', None) is not None:
             url += f'&replicaSet={self.connection_config["replica_set"]}'
 
+        if self.connection_config.get('ssl', None) is not None:
+            url += f'&ssl={self.connection_config["ssl"]}'
+
         return_code = subprocess.call([
             'mongodump',
             '--uri', f'"{url}"',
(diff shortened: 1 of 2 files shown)


pipelinewise/fastsync/commons/tap_mongodb.py
```diff
@@ -298,6 +298,9 @@ def _export_collection(self, export_dir: str, collection_name)->str:
         if self.connection_config.get('replica_set', None) is not None:
             url += f'&replicaSet={self.connection_config["replica_set"]}'
 
+        if self.connection_config.get('ssl', None) is not None:
+            url += f'&ssl={self.connection_config["ssl"]}'
```
koszti, line 302: `ssl` is an optional parameter and it might cause `KeyError: 'ssl'` at runtime.
Can we use `self.connection_config.get('ssl')` instead?

novotl: The case when `ssl` is not set is ruled out by the if branch. Similarly as with the `replica_set` parameter 2 lines above, when the option is not specified at all, the `url` is not modified.

koszti: ah apologies, I'm blind :(

Samira-El (approved):

PR #583, opened 2020-12-02
[AP-866] Tag snowflake queries

koszti:
## Problem

Tagging all snowflake queries with `tap_id`, `ppw_component`, `table` and `schema`. Query tags are writtin into Snowfalke query history table in JSON format.

Example output in query history:
image

## Proposed changes

1. Generating and utilising the `query_tag` optional argument in target-snowflake.
2. Fastsync to use `QUERY_TAG` session parameter, when running SQLs

## Types of changes

## Checklist


pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -60,12 +60,18 @@ def open_connection(self):
                                            autocommit=True,
                                            session_parameters={
                                                # Quoted identifiers should be case sensitive
-                                               'QUOTED_IDENTIFIERS_IGNORE_CASE': 'FALSE'
+                                               'QUOTED_IDENTIFIERS_IGNORE_CASE': 'FALSE',
+                                               'QUERY_TAG': json.dumps({
+                                                   'ppw_component': 'fastsync',
+                                                   'tap_id': self.connection_config.get('tap_id'),
+                                                   'table': query_tag_props.get('schema'),
```
Samira-El, line 67: `query_tag_props` can be `None`, calling `get` on it would fail.

Samira-El: Empty dictionary could be a better default value of `query_tag_props`

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -60,12 +60,18 @@ def open_connection(self):
...
+                                                   'schema': query_tag_props.get('table')
```
Samira-El, line 68: Looks like there is a typo here: `'schema': query_tag_props.get('table')` and above too.
`schema` value is being set with value of `table` key and vice versa.

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -120,14 +126,14 @@ def upload_to_s3(self, file, table, tmp_dir=None):
 
     def create_schema(self, schema):
         sql = 'CREATE SCHEMA IF NOT EXISTS {}'.format(schema)
-        self.query(sql)
+        self.query(sql, query_tag_props={'schema': schema, 'table': None})
```
Samira-El, line 129: Given how the `table` value is being dealt with in the `open_connection` method, there is no need to mention the table here `'table': None`.

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -195,21 +201,21 @@ def grant_select_on_table(self, target_schema, table_name, role, is_temporary, t
...
-            self.query(sql)
+            self.query(sql, query_tag_props={'schema': target_schema, 'table': table_name})
 
     # pylint: disable=unused-argument
     def grant_usage_on_schema(self, target_schema, role, to_group=False):
         # Grant role is not mandatory parameter, do nothing if not specified
         if role:
             sql = 'GRANT USAGE ON SCHEMA {} TO ROLE {}'.format(target_schema, role)
-            self.query(sql)
+            self.query(sql, query_tag_props={'schema': target_schema, 'table': None})
```
Samira-El, line 211: > Given how the table value is being dealt with in the open_connection method, there is no need to mention the table here 'table': None.

Same as this.

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -195,21 +201,21 @@ def grant_select_on_table(self, target_schema, table_name, role, is_temporary, t
...
 
     # pylint: disable=unused-argument
     def grant_select_on_schema(self, target_schema, role, to_group=False):
         # Grant role is not mandatory parameter, do nothing if not specified
         if role:
             sql = 'GRANT SELECT ON ALL TABLES IN SCHEMA {} TO ROLE {}'.format(target_schema, role)
-            self.query(sql)
+            self.query(sql, query_tag_props={'schema': target_schema, 'table': None})
```
Samira-El, line 218: same comment.

tests/units/cli/test_config.py
```diff
@@ -300,6 +300,9 @@ def test_save_config(self):
                 }
             },
             'temp_dir': './pipelinewise-test-config/tmp',
+            'tap_id': 'mysql_sample',
```
Samira-El, line 303: can you create another `test_config` and tests where the `query_tag` is not present?
This is to make sure the ppw and all component are backward compatible.

tests/units/cli/test_config.py
```diff
@@ -300,6 +300,9 @@ def test_save_config(self):
...
+            'query_tag':
```
Samira-El, line 304: can you create another test_config and tests where the `query_tag` is not present?
This is to make sure the ppw and all component are backward compatible.

pipelinewise/cli/config.py
```diff
@@ -318,6 +319,13 @@ def save_tap_jsons(self, target, tap, extra_config_keys=None):
         # Generate tap inheritable_config dict
         tap_inheritable_config = utils.delete_empty_keys({
             'temp_dir': self.get_temp_dir(),
+            'tap_id': tap.get('id'),
+            'query_tag': json.dumps({
```
Samira-El, line 323: How would ppw behave if the command `run_tap` is ran with config that doesn't have these new property?
From what I see, these changes are breaking and not backward compatible, unless all pipelines are stopped and their config are imported.

Samira-El (changes requested):
Left few comments

pipelinewise/cli/config.py
koszti: It's still working with the old config without these parameters, but not tagging the queries.

fastsync:
* `tap_id` and `query_tags` are optional values. If not defined then the query_tag will be generated with None values. Test added by 

target-snowflake:
* `query_tag` is an optional parameter in target-snowflake. If not defined in the `config.json` then nothing will be tagged by target-snowflake. [This test](https://github.com/transferwise/pipelinewise-target-snowflake/blob/195e7d7d9cd67f635f3c5c355dd7d31fc92740df/tests/integration/test_target_snowflake.py#L1046) is targeting when `query_tag` is defined in `config.json`. All the other test cases in the file running without `query_tag`

pipelinewise/fastsync/commons/target_snowflake.py
koszti: fixed by 26febe82084667ac7e7db47299147e0dfa284257

pipelinewise/fastsync/commons/target_snowflake.py
koszti: fixed by 26febe82084667ac7e7db47299147e0dfa284257

pipelinewise/fastsync/commons/target_snowflake.py
koszti: fixed by 26febe82084667ac7e7db47299147e0dfa284257

pipelinewise/fastsync/commons/target_snowflake.py
koszti: fixed by 26febe82084667ac7e7db47299147e0dfa284257

pipelinewise/fastsync/commons/target_snowflake.py
koszti: fixed by 26febe82084667ac7e7db47299147e0dfa284257

tests/units/cli/test_config.py
koszti: I can't create another test_config because it is generated and the expected output is to always include `tap_id` and `query_tag` in the generated config.

To make sure it's still compatible with old config which doesn't have these properties, please check the next point.

Samira-El: Got it

pipelinewise/cli/config.py
```diff
@@ -318,6 +319,13 @@ def save_tap_jsons(self, target, tap, extra_config_keys=None):
...
+                'ppw_component': tap.get('type'),
+                'tap_id': tap.get('id'),
+                'schema': '{schema}',
```
Samira-El, line 326: Is this supposed to be a f-string?

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -60,12 +73,13 @@ def open_connection(self):
                                            autocommit=True,
                                            session_parameters={
                                                # Quoted identifiers should be case sensitive
-                                               'QUOTED_IDENTIFIERS_IGNORE_CASE': 'FALSE'
+                                               'QUOTED_IDENTIFIERS_IGNORE_CASE': 'FALSE',
+                                               'QUERY_TAG': self.create_query_tag(query_tag_props)
```
Samira-El, line 77: Looking at the screenshots you shared in the PR description, the schema property is just the schema name only.
Do you think that maybe it would be problematic to have not have the DB included? what if we have multiple pipelines loading to the schemas with same name but in different DBs?

I think having either the full qualified schema name or DB also in the query tag would be useful if better filtering and analysing. We can get the db in this scope from `self.connection_config['dbname']`.
wdyt?

pipelinewise/cli/config.py
Samira-El: Asking because `'{schema}'` will output `'{schema}'`.

Samira-El: same with `table` below

Samira-El: Oh, I think this is meant to be a template or a placeholder, right?

Samira-El (commented):
left few smol comments.

pipelinewise/cli/config.py
koszti: yes, `query_tag` is a templated string where `{schema}` and `{table}` are tokens replaced dynamically at runtime by target-snowflake.

pipelinewise/fastsync/commons/target_snowflake.py
koszti: Voila.

Added to snowflake fastsync by 5672fb0409d24ef201729c9963872ffb2d118b3f
PPW passing the new template with database to target-snowflake by db01e64f2fd2d5da5f2f0451ea013bc3fdc44349
Target-snowflake now understands `{database}` token: https://github.com/transferwise/pipelinewise-target-snowflake/pull/117/files

pipelinewise/cli/config.py
Samira-El: Ok.

Samira-El: I'm a bit worried here, so in this [function](https://github.com/transferwise/pipelinewise/blob/25c13bbd0adbd0c3cb76dfd2b6a45e069c61005b/pipelinewise/cli/tap_properties.py#L93) in PPW, we use double curly brackets for templated strings `{{ token }}`, but here it's a single one.
I'm afraid we're not being consistent here!

koszti: Single brackets used in configurable string templates in target-s3-csv (`naming_convention` parameter) and in target-snowflake (`query_tag` parameter). So PPW targets components in sync.
 
Double brackets used in PPW in the above function to generate and not configurable.

What's the reason of the worries, what issues it can cause?
And which pattern we should use, double or single brackets? I'm afraid both cases requires changes in multiple places in multiple repositories ånd I think is out of the scope of this PR.

Samira-El: hmm, we don't want to beak anything.

I guess we can leave this for another PR.

koszti: no worries, I'll update the PRs to use jinja style double brackets for query tagging. Double brackets sounds like the correct option we should go with.

target-s3-csv will remain the same, and will use single brackets. We'll address that separately whenever it's needed. What do you think, is that a good plan?

Samira-El (approved):

PR #615, opened 2021-01-11
[AP-902][AP-905] Drop pg replication slot in case of full re-sync of a tap.

Samira-El:
## Problem

**AP-905** 
This is a bug in FastSync that allows running Fastsync with tables that aren't configured in the tap with the help of the arg `--tables`.

**AP-902**
When a full resync of the whole tap is triggered, PipelineWise creates a slot if it doesn’t exist.
while the full resync is taking place, if a slot already exists and occupies a lot of disk space, it will grow in size as it’s not consumed from, this can create disk space problems for small but active dbs. 

## Proposed changes

**AP-905** 
Check the list of tables provided via `tables` arg against the set of selected tables in the properties file of the tap. 
Change is in commit

**AP-902** 
Since Full resync is not making use of the slot anyway, and the latter might grow too much, add ability to drop the slot.
That is done automatically by checking the value of arg `tables`, if none is provided then enable the dropping slot by calling a class method `FastSyncTapPostgres.drop_slot`.

## Types of changes

## Checklist

### DIFF
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -3,8 +3,12 @@ import decimal
 import gzip
 import logging
 import re
+import sys
 import psycopg2
 import psycopg2.extras
+
+from typing import Dict
+
 
 from . import utils
 from ...utils import safe_column_name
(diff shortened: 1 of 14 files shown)


pipelinewise/fastsync/commons/tap_postgres.py
```diff
@@ -152,16 +241,7 @@ def fetch_current_log_pos(self):
...
-            self.connection_config['host'],
-            self.connection_config['port'],
-            self.connection_config['user'],
-            self.connection_config['password'],
-            self.connection_config['dbname']
-        )
-        self.primary_host_conn = psycopg2.connect(primary_host_conn_string)
-        # Set connection to autocommit
-        self.primary_host_conn.autocommit = True
+        self.primary_host_conn = self.get_connection(self.connection_config, prioritize_primary=True)
```
koszti, line 244: 🥇 , it connecting to pg was not only duplicated but `'ssl'` option was completely ignored on `primary_host_conn`. Not any more.

tests/units/fastsync/assertions.py
```diff
@@ -153,29 +158,35 @@ def assert_main_impl_should_exit_with_error_on_failure(main_impl: callable,
...
-                        mock_enter.return_value.map.assert_called_once()
+                    with patch(objects_to_mock.full_tap_class_nm) as tap_mock:
+                        tap_mock.return_value.drop_slot.side_effect = None
+
+                        ns = Namespace(**{
+                            'tables': ['table_1', 'table_2', 'table_3', 'table_4'],
+                            'target': 'sf',
+                            'transform': None,
+                            'drop_pg_slot': True,
+                            'tap': {},
```
koszti, line 169: what is the empty `tap` key is doing here?

Samira-El:
```
    if args.drop_pg_slot:
        FastSyncTapPostgres.drop_slot(args.tap)
```
It's to mock `args.tap` that's done when `drop_pg_slot` is `True`

koszti (approved):

koszti:
super nice

PR #619, opened 2021-01-15
AP-909 Customize number of cores to use for parallelisation in FastSync

Samira-El:
## Problem

FastSync currently uses the number of CPU cores as the size of processes pool to use to sync tables.
This means that this might overwhelm sources that cannot handle multiple processes trying to fetch data at the same time, e.g  databases with low IOPS.

## Proposed changes

Add ability to customize the size of the process pool. 

That is done by adding a source level property `fastsync_parallelism` in the pipelines config to set the size.

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

## Checklist


pipelinewise/fastsync/mongodb_to_postgres.py
```diff
@@ -110,7 +110,7 @@ def sync_table(table: str, args: Namespace) -> Union[bool, str]:
 def main_impl():
     """Main sync logic"""
     args = utils.parse_args(REQUIRED_CONFIG_KEYS)
-    cpu_cores = utils.get_cpu_cores()
+    cpu_cores = args.tap.get('fastsync_parallelism', utils.get_cpu_cores())
```
koszti, line 113: would be great to eliminate that so many duplicates in `x_to_y.py` files one day.

pipelinewise/cli/schemas/tap.json
```diff
@@ -22,6 +22,15 @@
         "target_schema"
       ]
     },
+    "db_conn": {
+      "type": "object",
+      "properties": {
+        "fastsync_parallelism": {
+          "type": "integer",
+          "minimum": 1
```
koszti, line 30: can we have a realistic max value here?

install.sh
```diff
@@ -59,7 +59,7 @@ make_virtualenv() {
     echo "Making Virtual Environment for [$1] in $VENV_DIR"
     python3 -m venv $VENV_DIR/$1
     source $VENV_DIR/$1/bin/activate
-    python3 -m pip install --upgrade pip setuptools
+    python3 -m pip install --upgrade pip setuptools wheel
```
koszti, line 62: not necessarily related to the topic in the PR, but ok

install.sh
```diff
@@ -74,8 +74,6 @@ make_virtualenv() {
     fi
 
     echo ""
-    echo "===== Checking dependencies for conflict ..."
-    python3 -m pip check && echo "No conflicts" || exit 1
```
koszti, line 78: same as above. Would be better to put it into a separated PR and keep the commit history cleaner, but we can squeeze it into this one as well I think.

pipelinewise/fastsync/mongodb_to_postgres.py
Samira-El: Indeed, so many files with same code!
Soon Peter, soon!

pipelinewise/cli/schemas/tap.json
Samira-El: Like what?
`multiprocessing.Pool` does accept 1 as the min size of the pool.

install.sh
Samira-El: True lol.
I don't want to make a separate PR for a small change, esp given how long it take for the CI to run.

install.sh
Samira-El: should I revert this change

pipelinewise/cli/schemas/tap.json
koszti: Setting `fastsync_parallelism` to 1000000 likely will cause a death on the host. Do we want to control it by setting a realistic max allowed value in the config, like 100 or we don't want to care and we trust the user and the OS?

install.sh
koszti: nah, just a note

pipelinewise/cli/schemas/tap.json
Samira-El: Wdyt about programmatically falling back to the number of available CPU cores if the given `fastsync_parallelism` is greater than the # cup cores?

koszti: sounds good to me

Samira-El: Made few updates:
* set maximum to 1000 ..we could re-consider it if anyone every uses a super powerful node and want to use 1000 cores. :P
* choose min between cpu cores and fastsync_parallelism

koszti (dismissed):

pipelinewise/cli/schemas/tap.json
koszti: 🙇

Samira-El:
@koszti Can you bless me with your seal of approval?

koszti (approved):

PR #697, opened 2021-04-21
Add option to disable fastsync

mhindery:
## Problem

We have encountered several times issues with fastsync. It is currently not possible to run without fastsync without editing source code (if you know where to look), and install from a customized source. We'd highly prefer having the easily accessible option to not use fastsync with a certain sync and use the 'normal' singer sync when we need to disable it for a certain table or scenario. While it is definitely faster using fastsync, it sometimes does not work correctly, and in those scenario's we'd like to continue our sync/ELT process without it until e.g. a new release contains a bugfix.

## Proposed changes

Add a cli flag to disable fastsync

## Types of changes

## Checklist

### DIFF
--- a/docs/user_guide/cli.rst
+++ b/docs/user_guide/cli.rst
@@ -50,6 +50,8 @@ Run a specific pipeline
 
 :--extra_log: Optional: Copy logging into PipelineWise logger to see tap run logs at runtime in STDOUT
 
+:--enable_fastsync: Optional: Use FastSync when possible. Enabled by default.
+
 
 .. _cli_stop_tap:
 
@@ -110,6 +112,8 @@ sync and resets the table bookmarks to their new location.
 :--tap: Tap connector id
 
 :--tables: Optional: Comma separated list of tables to sync from the data source.
+
+:--enable_fastsync: Optional: Use FastSync when possible. Enabled by default.
 
 
 .. _cli_import:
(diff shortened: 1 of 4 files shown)


koszti:
@mhindery I added a few changes and end-to-end tests on top of your PR at https://github.com/transferwise/pipelinewise/commit/457d9777795623a2a0492569c5bf39d561a85277 . Please not that I changed the `enabled_fastsync` to `disable_fastsync` command line option to keep it in sync with other CLI parameters.

But what is more important that I realised some issues with disabling fastsync in general:
* Resync should ignore the content of `state.json` and if `--tables` option added then it should ignore only the selected tables from the state file. For example running `pipelinewise sync_tables --tap <TAP_ID> --target <TARGET_ID> --tables my_schema.table1,my_schema.table2 --disable_fastsync` should ignore the state file for `table1` and `table2`. But calling [run_tap_singer](https://github.com/transferwise/pipelinewise/commit/457d9777795623a2a0492569c5bf39d561a85277#diff-d6db5c834e810ee64631e988b65df98b381d3c9104e0eaacc7a8e92abd5cf9fdL1192) is using the positions from the state file hence resync is not performed. 
* Some major steps in some tap types are done by fastsync and ignoring it can cause major issues. For example tap-postgres requires postgres to have a replication slot created in a well-defined format and name and this is created by fastsync. If we enable users to disable fastsync then these steps are completely ignored and could cause issues.


Maybe we should focus only on your other PR that fixes the mysql-to-postgres column type mapping? What do you think?

PR #786, opened 2021-08-25
AP-1010 Fix Decimal not JSON serializable in MongoDB FastSync

Samira-El:
## Problem

Mongodb taps which have Decimal fields fail with the error: `"Object of type 'Decimal' is not JSON serializable"`

This is because FastSync MongoDB uses a sub class of J`SONENCODER` where BSON’s `Decimal128` is serialized into Python’s built-in Decimal type, but then the Json Encoder still tries to serialize the latter into something but it doesn’t know what exactly because there is no mapping defined for Python’s Decimal type.

Tap-mongodb doesn’t use a subclass of JsonEncoder to serialize the record but rather just a function so it’s not running into the same issue.

Error log:
```
time=2021-05-07 08:53:41 logger_name=pipelinewise.fastsync.mongodb_to_snowflake log_level=ERROR message=Object of type 'Decimal' is not JSON serializable
Traceback (most recent call last):
File "/home/pipelinewise/pipelinewise/pipelinewise/fastsync/mongodb_to_snowflake.py", line 73, in sync_table
mongodb.copy_table(table, filepath, args.temp_dir)
File "/home/pipelinewise/pipelinewise/pipelinewise/fastsync/commons/tap_mongodb.py", line 175, in copy_table
'DOCUMENT': json.dumps(document, cls=MongoDBJsonEncoder, separators=(',', ':')),
File "/usr/lib/python3.6/json/_init_.py", line 238, in dumps
**kw).encode(obj)
File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
chunks = self.iterencode(o, _one_shot=True)
File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
return _iterencode(o, 0)
File "/home/pipelinewise/pipelinewise/pipelinewise/fastsync/commons/tap_mongodb.py", line 95, in default
return super().default(o)
File "/usr/lib/python3.6/json/encoder.py", line 180, in default
o._class.name_)
TypeError: Object of type 'Decimal' is not JSON serializable
```
## Proposed changes
Agreed upon solution is to stop using the Json encoder in FastSync and use a similar function as tap-mongodb to serialize the records.

I have included fixtures that include all supported datatypes to make sure all types can be serialized and have tests for that.

### Misc changes
* Installing all connectors slows down end-2-end tests, thus creating a new GHA job for that and only installing the necessary singer connectors e2e tests.
* Install mongo shell using ppa rather than downloading deb package. 

## Types of changes

## Checklist


pipelinewise/fastsync/commons/tap_mongodb.py
```diff
@@ -192,13 +232,17 @@ def copy_table(self,
...
-                        '_ID': str(document['_id']),
-                        'DOCUMENT': json.dumps(document, cls=MongoDBJsonEncoder, separators=(',', ':')),
-                        utils.SDC_EXTRACTED_AT: extracted_at,
-                        utils.SDC_BATCHED_AT: datetime.datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f'),
-                        utils.SDC_DELETED_AT: None
-                    })
+                    try:
+                        rows.append({
+                            '_ID': str(document['_id']),
+                            'DOCUMENT': ujson.dumps(serialize_document(document)),
```
koszti, line 238: why the new ujson? Is that for the extra performance reason or it's serialising some items differently?

Samira-El: it's the former, ujson is faster than the built in json module.

koszti: looks like a small effort task that we can do with all taps/targets. Do you know anything why we should not switch to ujson in every component? Obviously not part of this PR, just wondering.

koszti (approved):

pipelinewise/fastsync/commons/tap_mongodb.py
Samira-El: I wouldn't call it small effort task, it's indeed a small change but with the number of taps/target we have it's painful hahaha

PR #797, opened 2021-09-14
[AP-913][Part-1] FastSync to pull supported columns only - mysql/mariadb

Samira-El:
## Problem

FastSync currenctly when extracting data from sources, it pulls all columns even the ones with types unsupported by the equivalent taps making the FastSync and Singer not in sync.

## Proposed changes

__This PR is for Mysql/mariadb__

Add the list of supported column types in mysql/mariadb and centralize the mapping in one file to avoid repetition in every mysql -> target X file.

## Types of changes

## Checklist

### DIFF
--- a/pipelinewise/fastsync/commons/tap_mysql.py
+++ b/pipelinewise/fastsync/commons/tap_mysql.py
@@ -2,11 +2,14 @@ import csv
 import datetime
 import decimal
 import logging
+from typing import Dict, Callable
+
 import pymysql
 
 from pymysql import InterfaceError, OperationalError
 
 from . import utils, split_gzip
+from .type_mapping import SUPPORTED_MYSQL_DATATYPES
 from ...utils import safe_column_name
 
 LOGGER = logging.getLogger(__name__)
@@ -27,7 +30,7 @@ class FastSyncTapMySql:
     """
 
     def __init__(self, connection_config, tap_type_to_target_type, target_quote=None):
-        self.connection_config = connection_config
+        self.connection_config: Dict = connection_config
         self.connection_config['charset'] = connection_config.get(
             'charset', DEFAULT_CHARSET
         )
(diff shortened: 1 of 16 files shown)


koszti:
`VARCHAR`meant to be the fallback data type if no specific target data type found. It's `VARCHAR` because everything can be represented as string from a CSV file. Would that be ideally better to extend the tap(s) with more supported types and not removing an existing feature?

What will happen to the tables that's currently using `FULL_TABLE` if some columns currently using the  fallback `VARCHAR` type? Will it introduce a breaking change, so people will not see column(s), that they currently can? If this happens do we expect to improve the tap even if it's not used at all because it's configured as `FULL_TABLE`?

Samira-El:
>  Would that be ideally better to extend the tap(s) with more supported data types and not removing an existing feature?

well, it's not really a proper feature to begin with, fastsync would pull data from unsupported column but then singer replication would ignore those columns.

Extending the taps would an be ideal situation, perhaps something for the future since it requires much more work than fixing only fastsync. So for now, this is trying to align fastsync & singer.

Samira-El:
>    The mapper functions don't have a fallback data type but it returns `None`. What will happen to tables with these columns, will the entire sync fail or the columns will be ignored?


so `return None` should ideally be executed, because the source wouldn't even pull unsupported columns, if it does, then fastsync would fail.
I was debating maybe it's better to raise an exception that the column type is unsupported ignore the column when None is returned in the method where the mapped is used.

Samira-El:
> What will happen to the tables that's currently using FULL_TABLE with some columns currently using the fallback VARCHAR type? Will it introduce a breaking change, so people will not see column(s) or table(s), that they currently can?

Yes.

Samira-El:
> If people stop seeing column(s) as a result of FULL_TABLE from the previous point, then do we expect to improve the tap even if it's not used at all because it's configured as FULL_TABLE?

I would say yes, because at anytime we could switch to logbased or incremental and tap would start ignoring those columns anyway -> people would start seeing nulls in the column.

I'd rather not have the column in the target table than have it but be missing data.

Samira-El:
Most of the columns that are unsupported are of types which have no analytical value to a human like blobs and bytes, they're unreadable by humans.

PR #832, opened 2021-11-04
[AP-1056] Added cleanup method for state file

amofakhar:
## Problem

When launching `pipelinewise sync_tables` for a log_based table, it does not clean the `state.json` entries for the tables before it starts.
This means, that if sync_tables fails for a table, it will not retry on the next run, but instead switch to log_based

## Proposed changes

a clean up method is implemented to be run at startup and removing the state file or cleaning the tables from it (if running sync for specific tables)

## Types of changes

## Checklist


pipelinewise/cli/pipelinewise.py
```diff
@@ -1729,6 +1731,42 @@ def _run_post_import_tap_checks(
...
+            with open(state_file_to_clean, 'r+', encoding='UTF-8') as state_file:
+                state_data = json.load(state_file)
+                bookmarks = state_data.get('bookmarks')
+                list_of_tables = tables.split(',')
+                if bookmarks:
+                    cleanedup_bookmarks = {
+                        bookmark_name: bookmarks[bookmark_name]
+                        for bookmark_name in bookmarks
+                        if not any(table_name in bookmark_name for table_name in list_of_tables)
+                    }
```
Samira-El, line 1757: This could be written in a more efficient & simpler way:
```suggestion
               list_of_tables = tables.split(',')
                if bookmarks:
                    for table in list_of_tables:
                         bookmarks.pop(table_name, None)
```

pipelinewise/cli/pipelinewise.py
```diff
@@ -1729,6 +1731,42 @@ def _run_post_import_tap_checks(
...
+
+                    state_data['bookmarks'] = cleanedup_bookmarks
+
+                state_file.seek(0)
+                json.dump(state_data, state_file)
+                state_file.truncate()
```
Samira-El, line 1763: why do you need this?

tests/units/cli/test_cli.py
```diff
@@ -48,6 +52,42 @@ def teardown_method(self):
...
+        with open(test_state_file, 'w', encoding='UTF-8') as state_file:
+            json.dump(sample_state_data, state_file)
+
+    @staticmethod
+    def _assert_calling_sync_tables(pipelinewise: PipelineWise, side_effect_method: Optional[Callable] = None) -> None:
+        with patch('pipelinewise.cli.pipelinewise.PipelineWise.run_tap_fastsync') as mocked_fastsync:
+            if side_effect_method:
+                mocked_fastsync.side_effect = side_effect_method
+            pipelinewise.sync_tables()
+        mocked_fastsync.assert_called_once()
```
Samira-El, line 89: it is not advised to use `assert_called_once`, it's a fake method that will always pass. Instead use `assert mocked_fastsync.call_count == 1`

pipelinewise/cli/pipelinewise.py
amofakhar: I think your suggestion for line #1753 should be like this : `for table_name in list_of_tables:`
Any how, If you run `test_command_sync_tables_cleanup_state_if_file_exists_and_table_argument` after this suggestion, you will see it fails.
The reason: your simpler way only removes the bookmark if `table_name` is exactly same as bookmark key, but in reality it is not! the bookmark keys have some prefixes same as what is implemented in the tests.
i.e: bookmark in state file : `taped-table_name` but tables argument: `table_name`

tests/units/cli/test_cli.py
```diff
@@ -575,6 +615,54 @@ def test_command_sync_tables(self):
...
+    def test_command_sync_tables_cleanup_state_if_file_exists_and_no_table_argument(self):
+        """Testing sync_tables cleanup state if file exists and there is no table argument"""
+        def _assert_state_file_is_deleted(*args, **kwargs):
+            # pylint: disable=unused-argument
+            assert os.path.isfile(test_state_file) is False
+
+        pipelinewise = self._init_for_sync_tables_states_cleanup()
+        test_state_file = pipelinewise.tap['files']['state']
+        self._make_sample_state_file(test_state_file)
+        self._assert_calling_sync_tables(pipelinewise, _assert_state_file_is_deleted)
```
Samira-El, line 637: I don't understand why `_assert_state_file_is_deleted` is being used as the side effect of the `run_tap_fastsync` method

tests/units/cli/test_cli.py
```diff
@@ -575,6 +615,54 @@ def test_command_sync_tables(self):
...
+                }
+            }
+            with open(test_state_file, encoding='UTF-8') as state_file:
+                state_data = json.load(state_file)
+            assert state_data == expected_state_data
+
+        pipelinewise = self._init_for_sync_tables_states_cleanup(tables_arg='table1,table3')
+        test_state_file = pipelinewise.tap['files']['state']
+        self._make_sample_state_file(test_state_file)
+        self._assert_calling_sync_tables(pipelinewise, _assert_state_file_is_cleaned)
```
Samira-El, line 656: same comment as above

pipelinewise/cli/pipelinewise.py
amofakhar: this overwrites the cleaned up json in the file:
`state_file.seek(0)` : moving the pointer to the beginning of the file.
`json.dump(state_data, state_file)` : dumping the cleaned up `state_data` to the file.
`state_file.truncate()`: removing the remained bytes of the previous contenx

i.e:
 
 original content of the file = `footbar`
 new data = `test`
 
 if not seek(0) then new content of the file will be : `foobartest`
 if use seek(0) but not truncate, the the contnt of the file will be: `tester`

pipelinewise/cli/pipelinewise.py
Samira-El: Are you sure about that? the prefix in the bookmark keys is the DB/schema name, not the tap id, and when you run the `sync_table` command, you have to provide those as well, not just `table1,table2...`

So in this failing test case, the input is incorrect.

tests/units/cli/test_cli.py
amofakhar: This is the implementation of `assert_called_once`:

```python
   def assert_called_once(self):
        """assert that the mock was called only once.
        """
        if not self.call_count == 1:
            msg = ("Expected '%s' to have been called once. Called %s times.%s"
                   % (self._mock_name or 'mock',
                      self.call_count,
                      self._calls_repr()))
            raise AssertionError(msg)
```

How do you believe it is a fake method and always passes? 
To test it: just add a `return` in `pipelinewise.cli.pipelinewise.sync_tables` just before `self.run_tap_fastsync` and you see the test fails because of not calling the method.

tests/units/cli/test_cli.py
amofakhar: we need to be sure that cleaning the state file is done before calling the `self.run_tap_fastsync` 
if we don't check this priority, and only check that file is cleaned in the test, test can be passed but in reality it's possible cleaning up the file be done after running this method (by bad implementation or mistake or updating the code in the future).
By having this side effect for the mocked method, It assures the file is already cleaned when the method is called! because it clean up is checked inside the side effect method.

tests/units/cli/test_cli.py
amofakhar: same answer as above!

tests/units/cli/test_cli.py
Samira-El: https://engineeringblog.yelp.com/2015/02/assert_called_once-threat-or-menace.html

pipelinewise/cli/pipelinewise.py
Samira-El: got it, thanks!

pipelinewise/cli/pipelinewise.py
amofakhar: I checked again the tap codes, you are right. I will fix the test and implementation.

amofakhar: - [ ] Make the code simpler!

tests/units/cli/test_cli.py
amofakhar: 1. In this article they are talking about 'mock' package, in pipelinewise we use the mock from `unittest`. They are different!
1. This article is written on Feb 4 2015! on that time the latest version of `mock` was `1.0.1` which had this bug. the latest version of `mock` is '4.0.3' which this issue is solved.

Samira-El: hmmm ok

pipelinewise/cli/pipelinewise.py
```diff
@@ -1729,6 +1731,37 @@ def _run_post_import_tap_checks(
...
+    def _cleanup_tap_state_file(self) -> None:
+        tables = self.args.tables
+        state_file = self.tap['files']['state']
+        if tables:
+            self._clean_tables_from_bookmarks_in_state_file(state_file, tables)
+        else:
+            try:
+                os.remove(state_file)
+            except FileNotFoundError:
+                pass
```
Samira-El, line 1743: We have a helper function in `utils.py` called `silentremove` which you can use to remove files

tests/units/cli/test_cli.py
Samira-El: ok

Samira-El (approved):
left a comment.

In general, lgtm

Samira-El (approved):

PR #860, opened 2022-01-10
Update `pipelinewise-target-bigquery`

judahrand:
## Problem

The BigQuery target is out of sync with the latest version.

## Proposed changes

Update the BigQuery target.

## Types of changes

What types of changes does your code introduce to PipelineWise?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [x] Description above provides context of the change
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Unit tests for changes (not needed for documentation changes)
- [x] CI checks pass with my changes
- [x] Bumping version in `setup.py` is an individual PR and not mixed with feature or bugfix PRs
- [x] Commit message/PR title starts with `[AP-NNNN]` (if applicable. AP-NNNN = JIRA ID)
- [x] Branch name starts with `AP-NNN` (if applicable. AP-NNN = JIRA ID)
- [x] Commits follow "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"
- [x] Relevant documentation is updated including usage instructions

PR #861, opened 2022-01-10
Fix broken Postgres to BigQuery `fastsync`

judahrand:
## Problem

It seems that a comment was introduced on the wrong line in the Postgres `fastsync` module in the latest release. This is currently broken and needs a bug fix release urgently.

There was also a bug introduced that in `SplitGzipFile` if `mode` is binary `builtins.open` does not take an encoding... 

## Proposed changes

Move the comment.

Only include the encoding when relevant.

## Types of changes

## Checklist


judahrand:
@Samira-El

Samira-El:
Hey @judahrand , thanks for raising this PR!

looks like Pylint is unhappy.

judahrand:
> Hey @judahrand , thanks for raising this PR!
> 
> looks like Pylint is unhappy.

Yeah, it look like Pylint is suggesting the broken change... 🤦

judahrand:
@Samira-El The most recent changes should fix the linting error.

judahrand:
Well this is frustrating... Pretty sure pylint isn't complaining locally. I'll have another look at this now. Other than the linting are you happy to merge this PR?

Samira-El:
> Well this is frustrating... Pretty sure pylint isn't complaining locally. I'll have another look at this now. 

Do you have the same pylint version, and running the same command as CI?

> Other than the linting are you happy to merge this PR?

Would be great to have some unit tests for these little changes to avoid regression :) .

judahrand:
@Samira-El Right, Pylint definitely happy this time. I've also added a regression test for the bug in `SplitGzipFile`. The other one, however, was:
a) just obviously wrong and never should have gotten through review.
b) is actually already tested in `TestTargetBigquery` end to end test - unfortunately all the BigQuery tests seem to be skipped.
c) given that the failure occurs due to giving the SQL engine a bad query it is tricky to test as a unit test 

These reasons have prevented be adding a regression test for that change.

judahrand:
@Samira-El

judahrand:
@Samira-El

judahrand:
@Samira-El This now includes all code changes which are required to get the BigQuery e2e tests running again. #860 is also required since the current version is not compatible with Python 3.8. Can you please review and merge this?

judahrand:
You can see that these changes are included in our fork: https://github.com/thread/pipelinewise 

They pass the e2e tests. Without them the tests do not pass.

@Samira-El @jeet-parekh-wise

pipelinewise/fastsync/mongodb_to_bigquery.py
```diff
@@ -31,7 +31,7 @@
 LOCK = multiprocessing.Lock()
 
 
-def tap_type_to_target_type(mongo_type):
+def tap_type_to_target_type(mongo_type, *_):
```
Samira-El, line 34: why change the signature?

Samira-El: same with the other files

tests/end_to_end/test_target_bigquery.py
```diff
@@ -41,6 +41,8 @@ def setup_method(self):
         self.run_query_tap_postgres = self.e2e.run_query_tap_postgres
         self.run_query_target_bigquery = self.e2e.run_query_target_bigquery
         self.mongodb_con = self.e2e.get_tap_mongodb_connection()
+        self.e2e.setup_target_bigquery()
+        self.e2e.remove_all_state_files()
```
Samira-El, line 45: can this moved to the `teardown_method` instead?

pipelinewise/fastsync/mongodb_to_bigquery.py
judahrand: This makes all `tap_type_to_target_type`
functions are compatible with `assert_all_columns_exist`
by making them  all take a second parameter.

judahrand: I was seeing a bunch of failures in the tests due to the signature not matching here: https://github.com/transferwise/pipelinewise/blob/6934b3851512dbdd4280790bf253a0a13ab65684/tests/end_to_end/helpers/assertions.py#L346

tests/end_to_end/test_target_bigquery.py
judahrand: The state files need to be removed before a test runs not after. Eg. if I've run PipelineWise manually for some reason in dev then want to run the tests the state files need to go. Similarly for emptying the BigQuery datasets. 

Furthermore, removing this data on teardown is likely to make debugging failures more difficult (since the tests seems to exit on the first failure).

judahrand:
@Samira-El would it be possible for this to be merged before the next release?

judahrand:
@koszti I'm not sure how that test failure could be caused by my changes?

koszti:
> @koszti I'm not sure how that test failure could be caused by my changes?

I'm not sure but it may related to the thing how e2e tests are currently configured and it's triggered from a forked repository. @Samira-El what do you think, do you have any idea?

Samira-El:
> > @koszti I'm not sure how that test failure could be caused by my changes?
> 
> I'm not sure but it may related to the thing how e2e tests are currently configured and it doesn't like when it's triggered from a forked repository. @Samira-El what do you think, do you have any idea?

Yeah, github actions doesn't share the secrets with PRs from forked repos. So here, the SF env variables are empty.
Prolly need to be handled more elegantly by skipped the SF tests. will look into it.

Samira-El (approved):

Samira-El:
PR lgtm tho.

koszti (approved):

PR #902, opened 2022-02-20
Improve signal handling

judahrand:
## Problem

Only the `bash` process which PipelineWise creates needs to be
terminated. Any other processes are not PipelineWise's responsibility to
kill. Killing other processes can cause unexpected behaviour as these 
processes may pre-date PipelineWise being initiated.

In our case we launch GCSFuse before running PipelineWise using `exec`.
The current signal handler kills GCSFuse before killing the Singer process
which causes it to crash unexpectedly and result in a graceless exit.

## Proposed changes

Only kill the single process required.

## Types of changes

## Checklist


judahrand:
@Samira-El This seems like a pretty important bug. Took me a while to work out what was happening. We're running in K8s so doing initialization in an entry point script before `exec`ing PipelineWise. This bug meant that our Pods never exited gracefully and correctly renamed the log files meaning that they failed to restart.

jeet-parekh-wise:
@judahrand, sounds a bit odd that pipelinewise is terminating processes that it didn't start. And pipelinewise may have more than one child process. Not in a container environment though. Could you share the process tree with pids, names, and parent-pids from what you see inside the container?

If GCSFuse indeed appears as a child of Pipelinewise, then I think a better approach to terminate the child processes would be to check if `pipelinewise` appears in the name of the process. If yes, then terminate that process. Rather than terminating just the `children[-1]` process.

> In our case we launch GCSFuse before running PipelineWise using `exec`

Just to be clear, you mean "PipelineWise using exec", right? Not "GCSFuse using exec".
Does GCSFuse run as a background process (`&`)?

judahrand:
> @judahrand, sounds a bit odd that pipelinewise is terminating processes that it didn't start.
We wrap the PipelineWise `entrypoint.sh`found here in our own entrypoint script which does additional set up and then runs `exec /app/entrypoint.sh`. Thus any child processes of our entrypoint script are also children of PipelineWise and get terminated.

> And pipelinewise may have more than one child process. Not in a container environment though.
I understand that there can be grandchildren processes but how can the CLI generate more than one child process? It just calls out to `bash`. That `bash` child process then correctly propagates SIGTERM to its children.

> Could you share the process tree with pids, names, and parent-pids from what you see inside the container?
Unfortunately, now as we've fixed this issue in our fork and are running an updated version. 

> If GCSFuse indeed appears as a child of Pipelinewise, then I think a better approach to terminate the child processes would be to check if `pipelinewise` appears in the name of the process. If yes, then terminate that process. Rather than terminating just the `children[-1]` process.

This may work but feels like a hack. I still don't see how PipelineWise can have more than one child process using the CLI - which is the intended use case, right?

> Just to be clear, you mean "PipelineWise using exec", right? Not "GCSFuse using exec". Does GCSFuse run as a background process (`&`)?

I mean this:
```
gcsfuse --implicit-dirs bucket $PIPELINEWISE_CONFIG_DIRECTORY

# Run the entrypoint of the base PipelineWise image.
exec /app/entrypoint.sh "$@"
```

judahrand:
>  I think a better approach to terminate the child processes would be to check if pipelinewise appears in the name of the process.

Actually, this won't work. The child process's name is `/bin/bash` and the grandchildren are just `tap-foo` and `target-bar`. 

Additionally, looking at the entire command and checking for `pipelinewise` is asking for trouble and only processes may use the work in their args.

judahrand:
@jeet-parekh-wise 

Here is the output of `pstree`:
```
pstree
pipelinewise─┬─bash─┬─mbuffer───2*[{mbuffer}]
             │      ├─tap-postgres
             │      ├─target-bigquery
             │      └─transform-field
             └─gcsfuse───21*[{gcsfuse}]
```

jeet-parekh-wise:
Makes sense. I still wonder if `children[-1]` would be the right way to terminate the process. Is it guaranteed that it would always be the last process in the list?

judahrand:
> Makes sense. I still wonder if `children[-1]` would be the right way to terminate the process. Is it guaranteed that it would always be the last process in the list?

I believe the fact that it is last in the list is guaranteed as it will always be the youngest child process. But there may also be a way to sort by creation time to make sure.

judahrand:
@jeet-parekh-wise I added the sort based on creation time which should now guarantee that the youngest process is killed.

louis-pie:
> @jeet-parekh-wise
> 
> Here is the output of `pstree`:
> 
> ```
> pstree
> pipelinewise─┬─bash─┬─mbuffer───2*[{mbuffer}]
>              │      ├─tap-postgres
>              │      ├─target-bigquery
>              │      └─transform-field
>              └─gcsfuse───21*[{gcsfuse}]
> ```
> 
> You only need to send `SIGTERM` to the `bash` process as bash will correctly propagate the signal to its children.

Hi @judahrand ... how are you making pipelinewise.py launch gcsfuse ?

judahrand:
> > @jeet-parekh-wise
> > Here is the output of `pstree`:
> > ```
> > pstree
> > pipelinewise─┬─bash─┬─mbuffer───2*[{mbuffer}]
> >              │      ├─tap-postgres
> >              │      ├─target-bigquery
> >              │      └─transform-field
> >              └─gcsfuse───21*[{gcsfuse}]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You only need to send `SIGTERM` to the `bash` process as bash will correctly propagate the signal to its children.
> 
> Hi @judahrand ... how are you making pipelinewise.py launch gcsfuse ?

We aren't. We're wrapping the default entry point of the PipelineWise container in a our own entry point script which does this:

```
gcsfuse --implicit-dirs bucket $PIPELINEWISE_CONFIG_DIRECTORY

# Run the entrypoint of the base PipelineWise image.
exec /app/entrypoint.sh "$@"
```
This results in GCSFuse being a child process of the process which 'becomes' PipelineWise.

Does that make sense? PipelineWise isn't launching GCSFuse. The process which becomes PipelineWise (through exec) is launching GCSFuse.

louis-pie:
> > > @jeet-parekh-wise
> > > Here is the output of `pstree`:
> > > ```
> > > pstree
> > > pipelinewise─┬─bash─┬─mbuffer───2*[{mbuffer}]
> > >              │      ├─tap-postgres
> > >              │      ├─target-bigquery
> > >              │      └─transform-field
> > >              └─gcsfuse───21*[{gcsfuse}]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You only need to send `SIGTERM` to the `bash` process as bash will correctly propagate the signal to its children.
> > 
> > 
> > Hi @judahrand ... how are you making pipelinewise.py launch gcsfuse ?
> 
> We aren't. We're wrapping the default entry point of the PipelineWise container in a our own entry point script which does this:
> 
> ```
> gcsfuse --implicit-dirs bucket $PIPELINEWISE_CONFIG_DIRECTORY
> 
> # Run the entrypoint of the base PipelineWise image.
> exec /app/entrypoint.sh "$@"
> ```
> 
> where `/app/entrypoint.sh` is the default entry point. This results in GCSFuse being a child process of the process which 'becomes' PipelineWise.
> 
> Does that make sense? PipelineWise isn't launching GCSFuse. The process which becomes PipelineWise (through exec) is launching GCSFuse.

Sounds more like an issue with docker than with PipelineWise ... outside of docker, PipelineWise should take care to kill all it's children, would you not agree ?

judahrand:
> @judahrand, random thought, but could you get the pstree output with process group ids? `pstree -g`

```
pipelinewise(1)─┬─bash(1)─┬─mbuffer(1)─┬─{mbuffer}(1)
                │         │            └─{mbuffer}(1)
                │         ├─tap-postgres(1)
                │         ├─target-bigquery(1)─┬─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    ├─{target-bigquery}(1)
                │         │                    └─{target-bigquery}(1)
                │         └─transform-field(1)
                └─gcsfuse(43)─┬─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              ├─{gcsfuse}(43)
                              └─{gcsfuse}(43)
```
Interesting - this looks like it might be the 'right' thing to do. Kill all the processes in the same process group as PipelineWise.

judahrand:
> > > > @jeet-parekh-wise
> > > > Here is the output of `pstree`:
> > > > ```
> > > > pstree
> > > > pipelinewise─┬─bash─┬─mbuffer───2*[{mbuffer}]
> > > >              │      ├─tap-postgres
> > > >              │      ├─target-bigquery
> > > >              │      └─transform-field
> > > >              └─gcsfuse───21*[{gcsfuse}]
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > You only need to send `SIGTERM` to the `bash` process as bash will correctly propagate the signal to its children.
> > > 
> > > 
> > > Hi @judahrand ... how are you making pipelinewise.py launch gcsfuse ?
> > 
> > 
> > We aren't. We're wrapping the default entry point of the PipelineWise container in a our own entry point script which does this:
> > ```
> > gcsfuse --implicit-dirs bucket $PIPELINEWISE_CONFIG_DIRECTORY
> > 
> > # Run the entrypoint of the base PipelineWise image.
> > exec /app/entrypoint.sh "$@"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `/app/entrypoint.sh` is the default entry point. This results in GCSFuse being a child process of the process which 'becomes' PipelineWise.
> > Does that make sense? PipelineWise isn't launching GCSFuse. The process which becomes PipelineWise (through exec) is launching GCSFuse.
> 
> Sounds more like an issue with docker than with PipelineWise ... outside of docker, PipelineWise should take care to kill all it's children, would you not agree ?

I don't think this is a Docker issue (especially as we're not using Docker - GKE uses containerd 😉). I also don't think that PipelineWise should go around killing processes which it did not create, no. And in this case although PipleineWise is the parent process of GCSFuse the Python code did not create the child. All that the parent child relationship represents in this case is that GCSFuse was forked from the PID that PipelineWise now has.

jeet-parekh-wise:
> Interesting - this looks like it might be the 'right' thing to do. Kill all the processes in the same process group as PipelineWise.

I agree. I feel comfortable with this approach.

judahrand:
@jeet-parekh-wise I've pushed to implement this. I still forgone the `recursive=True` to get grandchildren based on the fact that the underlying call to `bash` should be able to hand the signal propagation more cleanly than we are. Is that okay?

judahrand:
Done.

jeet-parekh-wise:
LGTM

jeet-parekh-wise (approved):

jeet-parekh-wise:
Thanks @judahrand!

Issue #806, opened 2021-09-27
Pipelinewise bails when syncing empty tables, msg: No such file or directory: `..._fastsync_YDLKIZ9O.csv.gz`


gerardjp:
**Describe the bug**
When replicating tables between a tap (Postgres) and a target (also Postgres) with no records in them, pipelinewise/fastsync returns with the following error message for every empty table:

`No such file or directory: ... /root/.pipelinewise/tmp/pipelinewise___SNIP___20210927-144614-856033_fastsync_YDLKIZ9O.csv.gz`

**To Reproduce**

1. Have a running sync in place as described above (tap:PG > target:PG)
1. Add an empty table on the tap side
2. Run the replication again
4. It then fails to complete the replication altogether with a `No such file or directory` message per empty table.

**Expected behavior**
Complete the sync without erroring out.

**Your environment**
- Running this with the dockerized pipelinewise build. Tried it with latest and tag: v0.35.2
 - Source type Postgres v11
 - Target type Postgres v11

Source config:
```
---
id: "tap_accept"
name: "Accept tap test"
type: "tap-postgres"

db_conn:
  host: "20.20.20.21"
  port: 5432
  user: "postgres"
  password: "XXX"
  dbname: "tap_accept"
  logical_poll_total_seconds: 120

target: "target_accept"
batch_size_rows: 20000

schemas:
  - source_schema: "public"
    target_schema: "public"

    tables:
      - table_name: "androidappversion"
        replication_method: "FULL_TABLE"
    # ... SNIP ...
```

Target config:
```
---
id: "target_accept"
name: "Main Postgres base"
type: "target-postgres"

db_conn:
  host: "20.20.20.21"
  port: 5432
  user: "postgres"
  password: "XXX"
  dbname: "target_accept"
```
**Additional context**
- Above configs are censored (to avoid information disclosure) but accurate.
- I've seen this in a live setup and completely rebuild a local setup where I can reproduce problem.
- It breaks with FULL_TABLE as well as with LOG_BASED.
- When I remove all tables from the tap config that generate the error the sync works fine.

Samira-El:
Hey @gerardjp, this can happen, any reason you're syncing an empty table? usually, you would only want to configure tables which are populated.

gerardjp:
Hi Samira,

[...]

PR #909, opened 2022-03-04
AP-1072|Issue:806 FastSync Pg to Pg fails on empty tables

Samira-El:
## Problem

https://github.com/transferwise/pipelinewise/issues/806

The bookmark for incremental is also wrong, it's returning a null replication key value because there is no data.

## Proposed changes

Add condition on the existence of exported file, and only import into target PG if this file exists.
Return empty bookmarks for empty tables.

## Types of changes

## Checklist

### DIFF
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -351,7 +351,7 @@ class FastSyncTapPostgres:
             key_value = float(postgres_key_value)
 
         if key_value is None:
-            LOGGER.warning('Not replication value found for table %s, returning empty bookmark', table)
+            LOGGER.warning('No replication value found for table %s, returning empty bookmark', table)
             return {}
 
         return {
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -338,6 +338,11 @@ class FastSyncTapPostgres:
             )
 
         postgres_key_value = result[0].get('key_value')
+
+        if postgres_key_value is None:
+            LOGGER.warning('No replication value found for table %s, returning empty bookmark', table)
+            return {}
+
         key_value = postgres_key_value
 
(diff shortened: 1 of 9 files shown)


louis-pie (approved):

pipelinewise/fastsync/commons/tap_postgres.py
```diff
@@ -350,6 +350,10 @@ def fetch_current_incremental_key_pos(self, table, replication_key):
         elif isinstance(postgres_key_value, decimal.Decimal):
             key_value = float(postgres_key_value)
 
+        if key_value is None:
+            LOGGER.warning('No replication value found for table %s, returning empty bookmark', table)
+            return {}
+
```
amofakhar, line 356: It is better to move these lines between lines 341 and 344, exactly after getting 'key_value'. because in a case it is None, then it should pass `IF` clauses in lines 344, 347 and 350 and none of those if cases will be true. so by moving 'if key_value is None' before them, we will save some processing time.

pipelinewise/fastsync/commons/tap_postgres.py
```diff
@@ -332,7 +332,7 @@ def fetch_current_incremental_key_pos(self, table, replication_key):
         result = self.query(
             f'SELECT MAX({replication_key}) AS key_value FROM {schema_name}."{table_name}"'
         )
-        if len(result) == 0:
+        if not result:
```
amofakhar, line 335: I could not find any unit test for this method. It could be nice to implement a unit test to cover at least these new implementations

pipelinewise/fastsync/commons/tap_postgres.py
Samira-El: addressed

pipelinewise/fastsync/commons/tap_postgres.py
Samira-El: unit tests added

amofakhar (approved):
LGTM

PR #919, opened 2022-03-21
[AP-1104] Ignore log file status when deciding if a tap can be started

amofakhar:
## Problem

Since the implementation of a PID file, checking these log files should not longer be done.

## Proposed changes

Removed log file status checking to decide if a tap can be started

## Types of changes

## Checklist


jeet-parekh-wise (approved):

judahrand:
@amofakhar This will break distributed execution won't it? For example in Kubernetes the main process is always PID 1. So the PID in the `pid_file` will also always be 1. The PID file will therefore always identify the instance on PipelineWise which is running in the current Pod as the one who wrote the PID file (even if it didn't), check that the process name is correct (it will be) and continue running.

PR #929, opened 2022-04-02
Make sure that Postgres always get a consistent snapshot

### LINKED ISSUE #930: Tap Postgres FastSync inconsistent snapshot for logical replication
**Describe the bug**

I believe there is an edge case in the initial snapshot process for Postgres LOG_BASED replication.

This can only occur when snapshotting from a 'replica' and when the replication slot did not previously exist.

Under the above conditions it is possible for the following to occur:

1. a replication slot is created on the primary database
2. the most recent LSN of present on the replica is used as the starting point for logical replication
3. this LSN is actually before the replication slot was created / became consistent due to replication lag between the primary and replica
4. the replica is snapshotted but has missing data that will never be available to the replication slot

I believe this can be solved by #929 (WIP). This approach is inspired heavily by how Debezium does its initial snapshotting with the addition of waiting for the replica to catch up enough with the primary if needed.

**To Reproduce**
1. Difficultt

**Expected behavior**
I expect all rows in the source database to be present in the destination database.

**Screenshots**

**Your environment**
 - Version of PipelineWise, e.g branch/commit #/release/tag
 - Source type and setup
 - Target type and setup

**Additional context**

judahrand:
## Problem

It is currently possible for PipelineWise to select an LSN to stream from which will result in missing data once the initial snapshot has been conducted.

This is due to the current mechanism of just selecting either the most recent LSN on the primary or the most recent LSN which has been replayed on the replica: https://github.com/thread/pipelinewise/blob/cfd2c2091eb60a5b71ea38486594ccfb91a4fe1c/pipelinewise/fastsync/commons/tap_postgres.py#L297-L322

This is not a safe thing to do as it does not guarantee a consistent view of the database.

Fixes: #930

## Proposed changes

Inspired by how Debezium (another CDC tool written in Java) deals with initial snapshots of Postgres databases we will take advantage of two behaviours:

1. When a replica slot is created Postgres calculates the first LSN at which the slot will provide a consistent view of the database (it also sets `confirmed_flush_lsn` to this value). We will stream from that point forward but use the current state of the database as the snapshot.
2. Since PipelineWise only relieves logically decoded messages of complete transactions it should be safe to assume that any LSN that PipelineWise receives (and confirms) provides a consistent view of the database. Therefore, if the replication slot already exists we will take the current value of `confirmed_flush_lsn` to be the point at which we will start streaming.

In both of these cases, we may have data which will also be streamed to us by the replication slot but this should not be an issue as we are relying on an 'at least once' delivery mechanism already.

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc

### DIFF
--- a/pipelinewise/fastsync/commons/tap_postgres.py
+++ b/pipelinewise/fastsync/commons/tap_postgres.py
@@ -4,6 +4,7 @@ import logging
 import re
 import sys
 import textwrap
+import time
 import psycopg2
 import psycopg2.errors
 import psycopg2.extras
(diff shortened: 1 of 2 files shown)


judahrand:
@Samira-El This PR is ready for review, I believe.

judahrand:
@Samira-El

judahrand:
@Samira-El

PR #935, opened 2022-04-05
Revert "[AP-1104] Ignore log file status when deciding if a tap can be started"

judahrand:
Reverts transferwise/pipelinewise#919

@amofakhar This will break distributed execution won't it? For example in Kubernetes the main process is always PID 1. So the PID in the pid_file will also always be 1. The PID file will therefore always identify the instance of PipelineWise which is running in the current Pod as the one who wrote the PID file (even if it didn't), check that the process name is correct (it will be) and continue running.

The log extension based checking was better as it could not collide with anything. PIDs can collide when running in a distributed setting.

### DIFF
--- a/pipelinewise/cli/pipelinewise.py
+++ b/pipelinewise/cli/pipelinewise.py
@@ -1155,6 +1155,12 @@ class PipelineWise:
             self.logger.info('Tap %s is not enabled.', self.tap['name'])
             sys.exit(1)
 
+        # Run only if not running
+        tap_status = self.detect_tap_status(target_id, tap_id)
+        if tap_status['currentStatus'] == 'running':
+            self.logger.info('Tap %s is currently running.', self.tap['name'])
+            sys.exit(1)
+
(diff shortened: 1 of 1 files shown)


Samira-El:
Hi Judah, I think this shouldn't break distributed execution, I'm getting an error `log_level=ERROR message=Another instance of the tap is already running.` when a new tap run is triggered while the tap is still running, the `pidfile.PIDFile` will check if the given pid file exists and is running, if it's the case it would throw an exception `pidfile.AlreadyRunningError` which is handled.

I can't try on a k8s setup tho, would you be able to check that out yourself, thanks?

judahrand:
Ah, I think you're right - my mistake I'd read the logic [here](https://github.com/mosquito/python-pidfile/blob/d2ac227a5c3635db2541aa8d8e4507bdeb4b34e2/pidfile/pidfile.py#L15-L32) incorrectly, apologies.

PR #953, opened 2022-05-04
[AP-1214] Partial sync cli structure

amofakhar:
Structure for partial sync CLI

### DIFF
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -13,9 +13,10 @@ from datetime import datetime
 from typing import Optional, Tuple
 from pkg_resources import get_distribution
 
-from .utils import generate_random_string
-from .pipelinewise import PipelineWise
-from ..logger import Logger
+from pipelinewise.cli.utils import generate_random_string
+from pipelinewise.cli.pipelinewise import PipelineWise
+from pipelinewise.logger import Logger
+from pipelinewise.cli.errors import CommandSpecificArgumentsException
 
 __version__ = get_distribution('pipelinewise').version
 USER_HOME = os.path.expanduser('~')
@@ -143,6 +144,59 @@ def __disable_profiler(
         profiler.clear()
 
 
+def _validate_command_specific_arguments(args):
+    # Command specific argument validations
+    if args.command == 'init' and args.name == '*':
+        raise CommandSpecificArgumentsException('You must specify a project name using the argument --name')
+
+    if args.command in ['discover_tap', 'test_tap_connection', 'run_tap', 'stop_tap']:
+        if args.tap == '*':
+            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
+        if args.target == '*':
+            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
+
+    if args.command == 'sync_tables':
+        if args.tap == '*':
+            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
+        if args.target == '*':
+            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
+
+    if args.command == 'import_config':
+        if args.dir == '*':
+            raise CommandSpecificArgumentsException(
+                'You must specify a directory path with config YAML files using the argument --dir'
+            )
+
+    if args.command == 'validate' and args.dir == '*':
+        raise CommandSpecificArgumentsException(
+            'You must specify a directory path with config YAML files using the argument --dir'
+        )
+
+    if args.command == 'encrypt_string':
+        if not args.secret:
+            raise CommandSpecificArgumentsException(
+                'You must specify a path to a file with vault secret using the argument --secret'
+            )
+        if not args.string:
+            raise CommandSpecificArgumentsException('You must specify a string to encrypt using the argument --string')
+
+    if args.command == 'partial_sync_table':
+        if args.tap == '*':
+            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
+
+        if args.target == '*':
+            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
+
+        if args.table == '*':
+            raise CommandSpecificArgumentsException('You must specify a source table by using the argument --table')
+
+        if args.column == '*':
+            raise CommandSpecificArgumentsException('You must specify a column by using the argument --column')
+
+        if args.start_value == '*':
+            raise CommandSpecificArgumentsException('You must specify a start value by using the argument --start_value')
+
+
 # pylint: disable=too-many-branches,too-many-statements
 def main():
     """Main entry point"""
@@ -191,84 +245,25 @@ def main():
         action='store_true',
     )
     parser.add_argument('--table', type=str, default='*', help='Name of the table to partial sync')
-    parser.add_argument('--column', type=str, default='*', help='Name of the column to partial sync')
+    parser.add_argument('--column', type=str, default='*', help='Name of the column to use as sync key in partial sync')
     parser.add_argument('--start_value', type=str, default='*', help='start value of the column to partial sync')
     parser.add_argument('--end_value', type=str, default=False, help='end value of the column to partial sync')
 
     args = parser.parse_args()
 
-    # Command specific argument validations
-    if args.command == 'init' and args.name == '*':
-        print('You must specify a project name using the argument --name')
-        sys.exit(1)
-
-    if args.command in ['discover_tap', 'test_tap_connection', 'run_tap', 'stop_tap']:
-        if args.tap == '*':
-            print('You must specify a source name using the argument --tap')
-            sys.exit(1)
-        if args.target == '*':
-            print('You must specify a destination name using the argument --target')
-            sys.exit(1)
-
-    if args.command == 'sync_tables':
-        if args.tap == '*':
-            print('You must specify a source name using the argument --tap')
-            sys.exit(1)
-        if args.target == '*':
-            print('You must specify a destination name using the argument --target')
-            sys.exit(1)
-
     # import and import_config commands are synonyms
     #
     # import        : short CLI command name to import project
     # import_config : this is for backward compatibility; use 'import' instead from CLI
-    if args.command == 'import' or args.command == 'import_config':
-        if args.dir == '*':
-            print(
-                'You must specify a directory path with config YAML files using the argument --dir'
-            )
-            sys.exit(1)
-
-        # Every command argument is mapped to a python function with the same name, but 'import' is a
-        # python keyword and can't be used as function name
+    # Every command argument is mapped to a python function with the same name, but 'import' is a
+    # python keyword and can't be used as function name
+    if args.command == 'import':
         args.command = 'import_project'
-
-    if args.command == 'validate' and args.dir == '*':
-        print(
-            'You must specify a directory path with config YAML files using the argument --dir'
-        )
-        sys.exit(1)
-
-    if args.command == 'encrypt_string':
-        if not args.secret:
-            print(
-                'You must specify a path to a file with vault secret using the argument --secret'
-            )
-            sys.exit(1)
-        if not args.string:
-            print('You must specify a string to encrypt using the argument --string')
-            sys.exit(1)
-
-    if args.command == 'partial_sync_table':
-        if args.tap == '*':
-            print('You must specify a source name using the argument --tap')
-            sys.exit(1)
-
-        if args.target == '*':
-            print('You must specify a destination name using the argument --target')
-            sys.exit(1)
-
-        if args.table == '*':
-            print('You must specify a source table by using the argument --table')
-            sys.exit(1)
-
-        if args.column == '*':
-            print('You must specify a column by using the argument --column')
-            sys.exit(1)
-
-        if args.start_value == '*':
-            print('You must specify a start value by using the argument --start_value')
-            sys.exit(1)
+    try:
+        _validate_command_specific_arguments(args)
+    except CommandSpecificArgumentsException as e:
+        print(str(e))
+        exit(1)
 
     logger = __init_logger(args.log, args.debug)
 
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -39,6 +39,7 @@ COMMANDS = [
     'import_config',  # This is for backward compatibility; use 'import' instead
     'validate',
     'encrypt_string',
+    'partial_sync_table'
 ]
 
 
@@ -189,6 +190,10 @@ def main():
         'The stats will be dumped into a folder in .pipelinewise/profiling',
         action='store_true',
     )
+    parser.add_argument('--table', type=str, default='*', help='Name of the table to partial sync')
+    parser.add_argument('--column', type=str, default='*', help='Name of the column to partial sync')
+    parser.add_argument('--start_value', type=str, default='*', help='start value of the column to partial sync')
+    parser.add_argument('--end_value', type=str, default=False, help='end value of the column to partial sync')
 
     args = parser.parse_args()
 
@@ -242,6 +247,27 @@ def main():
             sys.exit(1)
         if not args.string:
             print('You must specify a string to encrypt using the argument --string')
+            sys.exit(1)
+
+    if args.command == 'partial_sync_table':
+        if args.tap == '*':
+            print('You must specify a source name using the argument --tap')
+            sys.exit(1)
+
+        if args.target == '*':
+            print('You must specify a destination name using the argument --target')
+            sys.exit(1)
+
+        if args.table == '*':
+            print('You must specify a source table by using the argument --table')
+            sys.exit(1)
+
+        if args.column == '*':
+            print('You must specify a column by using the argument --column')
+            sys.exit(1)
+
+        if args.start_value == '*':
+            print('You must specify a start value by using the argument --start_value')
             sys.exit(1)
 
     logger = __init_logger(args.log, args.debug)
--- a/pipelinewise/cli/__init__.py
+++ b/pipelinewise/cli/__init__.py
@@ -149,13 +149,7 @@ def _validate_command_specific_arguments(args):
     if args.command == 'init' and args.name == '*':
         raise CommandSpecificArgumentsException('You must specify a project name using the argument --name')
 
-    if args.command in ['discover_tap', 'test_tap_connection', 'run_tap', 'stop_tap']:
-        if args.tap == '*':
-            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
-        if args.target == '*':
-            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
-
-    if args.command == 'sync_tables':
+    if args.command in ['discover_tap', 'test_tap_connection', 'run_tap', 'stop_tap', 'sync_tables']:
         if args.tap == '*':
             raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
         if args.target == '*':
@@ -194,7 +188,8 @@ def _validate_command_specific_arguments(args):
             raise CommandSpecificArgumentsException('You must specify a column by using the argument --column')
 
         if args.start_value == '*':
-            raise CommandSpecificArgumentsException('You must specify a start value by using the argument --start_value')
+            raise CommandSpecificArgumentsException(
+                'You must specify a start value by using the argument --start_value')
 
 
 # pylint: disable=too-many-branches,too-many-statements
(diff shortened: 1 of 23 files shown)


pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+        target_type = self.target['type']
+        target_id = self.target['id']
+
+        if target_type != 'target-snowflake':
+            print(f'Error! target "{target_id} (type: {target_type})" is not supported for partial sync!')
+            sys.exit(1)
+
+        if tap_type not in ['tap-mysql', 'tap-postgres']:
+            print(f'Error! tap "{tap_id} (type: {tap_type})" is not supported for partial sync!')
+            sys.exit(1)
```
Samira-El, line 1768: This will be hard to maintain as we support more tap/target types, we might end up with some unsupported tap-target combo, that happened before with FastSync. I suggest using something like https://github.com/transferwise/pipelinewise/blob/d9930bc9d586c292de763e23276806d4aa09519a/pipelinewise/cli/pipelinewise.py#L29

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+                tap_type,
+                target_type
+            )
+            sys.exit(1)
+
+    def _check_if_tap_is_enabled(self):
+        # Run only if tap enabled
+        if not self.tap.get('enabled', False):
+            self.logger.info('Tap %s is not enabled.', self.tap['name'])
+            sys.exit(1)
```
Samira-El, line 1784: I suggest raising an exception.

setup.py
```diff
@@ -69,6 +69,7 @@
               'mongodb-to-snowflake=pipelinewise.fastsync.mongodb_to_snowflake:main',
               'mongodb-to-postgres=pipelinewise.fastsync.mongodb_to_postgres:main',
               'mongodb-to-bigquery=pipelinewise.fastsync.mongodb_to_bigquery:main',
+              'partial-mysql-to-snowflake=pipelinewise.fastsync.partialsync.mysql_to_snowflake:main'
```
Samira-El, line 72: `pipelinewise.fastsync.partialsync.mysql_to_snowflake` module doesn't exist yet, I suggest removing it from this PR.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1072,6 +1072,50 @@ def update_state_file_with_extra_log(line: str) -> str:
...
+            end_value=self.args.end_value
+        )
+
+
+        # Do not run if another instance is already running
+        log_dir = os.path.dirname(self.tap_run_log_file)
+        if (
+                os.path.isdir(log_dir)
+                and len(utils.search_files(log_dir, patterns=['*.log.running'])) > 0
+        ):
```
Samira-El, line 1097: I believe there was a done ticket where we have decided not to use log files as a way to check if tap is running, It think this should also apply to this new command. The existence, or not, of pid file should be enough.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+                    target_id=target_id,
+                )
+
+                self.run_tap_partialsync(
+                    tap=tap_params, target=target_params, transform=transform_params,
+                )
+
+        except pidfile.AlreadyRunningError:
+            self.logger.error('Another instance of the tap is already running.')
+            sys.exit(1)
```
Samira-El, line 1743: Please raise an exception

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+        # Delete temp file if there is any
+        except commands.RunCommandException as exc:
+            self.logger.exception(exc)
+            self.send_alert(message=f'Failed to sync tables in {tap_id} tap', exc=exc)
+            sys.exit(1)
```
Samira-El, line 1748: Please raise an exception

pipelinewise/cli/pipelinewise.py
```diff
@@ -1072,6 +1072,50 @@ def update_state_file_with_extra_log(line: str) -> str:
             with open(tap.state, 'w', encoding='utf-8') as statefile:
                 statefile.write(state)
 
+    def run_tap_partialsync(self, tap: TapParams, target: TargetParams, transform: TransformParams):
+        """Running the tap for partial sync table"""
+
+        # Build the partial sync executable command
+        command = commands.build_fastsync_partial_command(
```
Samira-El, line 1079: Before running this command, can we have some validation that the given table and column exist and selected? 
We can do that by checking the properties file.

Samira-El: Furthermore, we need some validation of the column type, not all column types are suitable to use, I don't think a boolean or string or binary...etc column be useful.

pipelinewise/cli/__init__.py
```diff
@@ -189,6 +190,10 @@ def main():
         'The stats will be dumped into a folder in .pipelinewise/profiling',
         action='store_true',
     )
+    parser.add_argument('--table', type=str, default='*', help='Name of the table to partial sync')
+    parser.add_argument('--column', type=str, default='*', help='Name of the column to partial sync')
+    parser.add_argument('--start_value', type=str, default='*', help='start value of the column to partial sync')
+    parser.add_argument('--end_value', type=str, default=False, help='end value of the column to partial sync')
```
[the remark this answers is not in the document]
amofakhar: - [ ] Fix the help message

pipelinewise/cli/__init__.py
```diff
@@ -244,6 +249,27 @@ def main():
             print('You must specify a string to encrypt using the argument --string')
             sys.exit(1)
 
+    if args.command == 'partial_sync_table':
+        if args.tap == '*':
+            print('You must specify a source name using the argument --tap')
+            sys.exit(1)
```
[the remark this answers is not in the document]
amofakhar: - [ ] Refactor args validation

pipelinewise/cli/commands.py
```diff
@@ -354,6 +354,48 @@ def build_singer_command(
...
+                    f'--transform {transform.config}'
+                    if transform.config and os.path.isfile(transform.config)
+                    else '',
+                    f'--table {table}',
+                    f'--column {column}',
+                    f'--start_value {start_value}',
+                    f'--end_value {end_value}'
+                ],
+            )
+        )
```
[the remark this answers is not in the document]
amofakhar: This can be None:

```python
f'--transform {transform.config}' if transform.config and os.path.isfile(transform.config) else '',
```

setup.py
amofakhar: - [ ] Remove it!

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Refactor supporting check!

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+
+    def _check_if_complete_tap_configuration(self, fastsync_bin, tap_type, target_type):
+        # Tap exists but configuration not completed
+        if not os.path.isfile(fastsync_bin):
+            self.logger.error(
+                'Table sync function is not implemented from %s datasources to %s type of targets',
+                tap_type,
+                target_type
+            )
+            sys.exit(1)
```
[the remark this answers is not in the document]
amofakhar: - [ ] Fix it!

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1655,141 @@ def encrypt_string(self):
...
+        if not self.tap.get('enabled', False):
+            self.logger.info('Tap %s is not enabled.', self.tap['name'])
+            sys.exit(1)
+
+    def _check_if_state_exists(self):
+        state_file = self.tap["files"]["state"]
+        if os.path.exists(state_file):
+            return
+        self.logger.error('Could not find state file in "%s"!', state_file)
+        sys.exit(1)
```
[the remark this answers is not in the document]
amofakhar: - [ ] Fix it!

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Fix it!

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Fix it!

pipelinewise/cli/pipelinewise.py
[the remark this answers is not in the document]
amofakhar: These `sys.exit(1)`'s are not something new and they existed there. since it is not in this PR scope and also I worry if it causes some tests/functions break, I prefer not to touch them

pipelinewise/cli/pipelinewise.py
amofakhar: answered!

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] fix it

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Add validation for table and column

amofakhar: I could not find out how to know the column type from `properties.json` do you know any way for that? 
I also was thinking it can be handled in the sync module when it is going to run the query and if the selected column be an invalid then it would fail and return the error message in that stage. What do you think?

pipelinewise/cli/commands.py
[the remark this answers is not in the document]
Samira-El: No, it will either be `"--transform xyz"` or `""`, empty string is not None.

pipelinewise/cli/pipelinewise.py
Samira-El: Properties file has the json schema of the table (see `schema` key), with the properties (i.e columns) and their type.

Samira-El: I reckon we should fail fast, not until the sync starts and query errors.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1448,11 +1482,14 @@ def sync_tables(self):
             self.logger.exception(exc)
             self.send_alert(message=f'Failed to sync tables in {tap_id} tap', exc=exc)
             sys.exit(1)
+        except PreRunChecksException:
+            sys.exit(1)
```
Samira-El, line 1486: I suggesting Re-raising the error rather than exiting with 1

pipelinewise/cli/pipelinewise.py
```diff
@@ -1448,11 +1482,14 @@ def sync_tables(self):
...
         except Exception as exc:
             self.send_alert(message=f'Failed to sync tables in {tap_id} tap', exc=exc)
             raise exc
         finally:
-            utils.silentremove(cons_target_config)
+            if cons_target_config:
```
Samira-El, line 1491: I don't think this is needed.
`utils.silentremove` function would not error if the the given file doesn't exist, hence the `silent` in the name

pipelinewise/cli/pipelinewise.py
Samira-El: Yes I know, please re-raise it, the program will exit with non-zero code anyway and we will get better error logs than just `SystemExit`

pipelinewise/cli/pipelinewise.py
Samira-El: same as above

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Raise exception

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Raise Exception

pipelinewise/cli/pipelinewise.py
amofakhar: It will, if its input parameter be None. there are some cases that checks can be failed before defining value for `cons_target_config`

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1658,168 @@ def encrypt_string(self):
...
+        table_in_properties = next(
+            (item for item in streams if item['table_name'] == table_name), None
+        )
+        if table_in_properties is None:
+            self.logger.error('Not found table "%s" in properties!', table_name)
+            raise PreRunChecksException()
+
+        try:
+            table_in_properties['schema']['properties'][self.args.column]
+        except Exception as exp:
```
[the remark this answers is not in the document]
amofakhar: - [ ] catch KeyError

pipelinewise/cli/pipelinewise.py
```diff
@@ -52,6 +56,15 @@
...
 
+PARTIAL_SYNC_PAIRS = {
+    ConnectorType.TAP_MYSQL: {
+        ConnectorType.TARGET_SNOWFLAKE
+    },
+    ConnectorType.TARGET_POSTGRES: {
+        ConnectorType.TARGET_SNOWFLAKE
+    }
+
+}
```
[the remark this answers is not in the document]
amofakhar: Yes that's true. they will come in the next PR's

pipelinewise/cli/pipelinewise.py
jeet-parekh-wise: The other CLI commands - `run_tap`, `stop_tap`, `sync_tables` - they all exit with `sys.exit(1)` in case of a known error. Here, it's easy to figure out that the tap is already running. So better to keep it consistent and not raise an exception.

pipelinewise/cli/pipelinewise.py
amofakhar: I think `type` is the type of json object and I can see for example for `datetime` fields its type is `string` but `datetime` should be allowed. I was thinking maybe can use `format` field. but I see it is not the thing be included for all objects. 
example: 

```json
    "properties": {
        "date_created": {
            "format": "date-time",
            "inclusion": "available",
            "type": ["null", "string"]
         },
        "date_updated": {
            "format": "date-time",
            "inclusion": "available",
            "type": ["null", "string"]
        },
        "isActive": {
            "inclusion": "available",
            "type": ["null", "boolean"]
        },
        "original_date_created": {
            "inclusion": "available",
            "maxLength": 50,
            "type": ["null", "string"]
         },
        "weight_unit_id": {
            "inclusion": "automatic",
            "maximum": 127,
            "minimum": -128,
            "type": ["null", "integer"]
}
}
```

pipelinewise/cli/commands.py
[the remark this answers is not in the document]
Samira-El: Wow, that's true!
my bad then.

pipelinewise/cli/pipelinewise.py
Samira-El: Checking both `type` and `format` (if it exist) should be enough then to make a decision on the validity of the column as a sync key.

pipelinewise/cli/pipelinewise.py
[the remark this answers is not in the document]
Samira-El: Ok

pipelinewise/cli/pipelinewise.py
amofakhar: - [ ] Add checks for type

pipelinewise/cli/__init__.py
```diff
@@ -142,6 +144,59 @@ def __disable_profiler(
...
+        if args.tap == '*':
+            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
+        if args.target == '*':
+            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
+
+    if args.command == 'sync_tables':
+        if args.tap == '*':
+            raise CommandSpecificArgumentsException('You must specify a source name using the argument --tap')
+        if args.target == '*':
+            raise CommandSpecificArgumentsException('You must specify a destination name using the argument --target')
```
Samira-El, line 162: the `if args.command == 'sync_tables':` can be merged with the first if.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1658,171 @@ def encrypt_string(self):
...
+
+        # Continue only if tap and target is supported by partial sync
+        try:
+            self._check_supporting_tap_and_target_for_partial_sync()
+
+            tap_id = self.tap['id']
+            tap_type = self.tap['type']
+            target_id = self.target['id']
+            target_type = self.target['type']
+            fastsync_bin = utils.get_fastsync_bin(self.venv_dir, tap_type, target_type)
```
Samira-El, line 1676: `get_fastsync_bin` in partial sync ?

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1658,171 @@ def encrypt_string(self):
...
+
+    def _check_if_complete_tap_configuration(self, fastsync_bin, tap_type, target_type):
+        # Tap exists but configuration not completed
+        if not os.path.isfile(fastsync_bin):
+            self.logger.error(
+                'Table sync function is not implemented from %s datasources to %s type of targets',
+                tap_type,
+                target_type
+            )
+            raise SystemExit(1)
```
Samira-El, line 1787: I suggest raising a `NotImplementedError`, it's more proper for this check.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1621,6 +1658,171 @@ def encrypt_string(self):
...
+        if not self.tap.get('enabled', False):
+            self.logger.info('Tap %s is not enabled.', self.tap['name'])
+            raise PreRunChecksException()
+
+    def _check_if_state_exists(self):
+        state_file = self.tap['files']['state']
+        if os.path.exists(state_file):
+            return
+        self.logger.error('Could not find state file in "%s"!', state_file)
+        raise PreRunChecksException()
```
Samira-El, line 1800: Would it be a problem if the state file exists but it's empty or a certain stream has no bookmark?

tests/units/cli/resources/test_partial_sync/target_snowflake/tap_no_state/log/target_snowflake-tap_no_state-20220503_151213.partialsync.log.failed
```diff
@@ -0,0 +1 @@
+/bin/bash: ./virtualenvs-dummy/pipelinewise/bin/partial-mysql-to-snowflake: No such file or directory
```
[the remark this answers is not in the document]
amofakhar: It was added by mistake!

- [ ] Delete the file!

pipelinewise/cli/__init__.py
amofakhar: - [ ] Merge if statements

pipelinewise/cli/pipelinewise.py
amofakhar: They use same method for getting the bin. I rename the variable.

- [ ] rename `fastsync_bin` variable

pipelinewise/cli/pipelinewise.py
amofakhar: These parts of the codes are just copied from the existing codes. raising different exception causes some other tests that are not related to this PR to be failed because it is used by other parts of the project. here the main big method is only broken into some small methods to be more clean.
Spending time to refactor the other parts that makes also some concerns which may cause to break some functionalities is out of scope.

pipelinewise/cli/pipelinewise.py
amofakhar: there are two cases:
A) there is end_value: state file won't be updated and file existence means it was run before. that would be enough for running partial sync even if file is empty or damaged.
B) no end_value: state file will be updated. I trusted on the current existing method for updating state file. if it fails or can not update empty or damaged file then it can be a bug which needs its own tests and fixing in a separate task.

pipelinewise/cli/pipelinewise.py
```diff
@@ -1673,7 +1673,7 @@ def partial_sync_table(self):
             tap_type = self.tap['type']
             target_id = self.target['id']
             target_type = self.target['type']
-            fastsync_bin = utils.get_fastsync_bin(self.venv_dir, tap_type, target_type)
+            sync_bin = utils.get_fastsync_bin(self.venv_dir, tap_type, target_type)
```
Samira-El, line 1676: the variable name is not the problem, the wrong function is being called.

PR #979, opened 2022-06-24
[AP-1149] Partial Sync

amofakhar:
This PR includes all Partial Sync sub tasks which are reviewed before. 

new things in this PR:

- e2e tests 
- documentation


podviaznikov:
what was the initial use case for partial sync, would like to read back story

PR #1014, opened 2022-09-09
[AP-1255] Fixed partial sync if table does not exist in target

amofakhar:
## Problem

Partial sync was failing if table does not exist in the target

## Proposed changes

creating the table in the target if it does not exist

## Types of changes

## Checklist

- [] Relevant documentation is updated including usage instructions


pipelinewise/fastsync/partialsync/mysql_to_snowflake.py
```diff
@@ -43,6 +42,18 @@ def partial_sync_table(args: Namespace) -> Union[bool, str]:
...
+
+        # making target table if not exists
+        snowflake.create_table(
+            target_schema=target_schema,
+            table_name=target_table,
+            columns=snowflake_types['columns'],
+            primary_key=snowflake_types.get('primary_key'),
+            is_temporary=False,
+            sort_columns=False,
+            if_not_exists=True
```
louis-pie, line 54: I find this parameter slightly confusing because it is True on a negative, and also default **False**

Could we change it to be something that is True on a positive, like `table_already_exists` and rework the logic around that ?

amofakhar: - [ ] Fix it!

louis-pie (approved):

PR #1024, opened 2022-10-06
Upgrade the locking system

judahrand:
## Problem

We have been facing deadlocks. This is because the locks held by `pidfile` never expire. So if a process is `SIGKILL`ed (for example by Kubernetes) this will cause a deadlock.

## Proposed changes

This alternative locking library `sherlock` allows us to set a file based lock which is similar to the `pidfile`. However, it has a number of advantages:

- The locks expire. Therefore, we are able to hold the lock for only a short time (configured to 30 seconds) and renew it whenever it is close to expiring. If the process is ever killed the lock will expire after at most 30 seconds.
- The 'owner' of the lock is a UUID rather than a PID. This has advantages when running in Kubernetes since in Kubernetes the PID is ALWAYS 1. Therefore all instances think they have the lock. This was previously okay as PipelineWise also checked the log file for a `.running` extension. With the new locking this is no longer needed and has been removed.

## Types of changes

## Checklist


judahrand:
@Samira-El This locking mechanism is much much more robust than the PIDFile system used previously. Can you please have a look? I think this is a seriously good addition.

PR #1058, opened 2023-01-13
AP-1106: Cleanup after deleted taps and targets

Samira-El:
## Problem

Currently, when the yaml config of a tap or target gets removed, all the runtime files of deleted config remains on disk, and for postgres taps that used log_based, a replication slot will remain on the DB which will grow and cause disk space issues to the PG db. 

## Proposed changes

Delete all the runtime files of deleted taps and targets and drop replication slot from PG dbs if it exists.
This will be done during the `import` time of the yaml config files and will make use of PPW's generated `config.json`. So when the `pipelinewise import` command is invoked: 

1. Make a copy of the existing `config.json` in memory.
2. Current import logic - no changes -> this will generate a new `config.json`
3. compare old config with new and detect any deleted taps and targets, for each remove their folders on disk and if any tap is tap-postgres, connect to the source DB and drop replication slot if exists.

## Types of changes

## Checklist


louis-pie:
Ticket only asks that the replication slot be dropped from the source.

This change goes further than the ticket asks for by cleaning up all the configuration files as well ... this wasn't ~~groomed~~ refined. Are we sure this is what we want ?

louis-pie (approved):

PR #1279, opened 2026-06-05
tap-postgres: discover ltree and interval columns as strings

cameron-process-st:
## Context

We use PipelineWise's tap-postgres to replicate Postgres tables that include `ltree` (hierarchical label paths) and `interval` columns. During discovery, `schema_for_column_datatype` has no branch for either type, so it returns an empty schema and those columns reach the target untyped (effectively dropped). This adds explicit handling so they're discovered as strings.

## What

In `schema_for_column_datatype`, map `interval` and `ltree` to a nullable string schema (non-nullable for primary keys, via `nullable_column`) — consistent with how `uuid`, `citext`, and `cidr/inet/macaddr` are already handled. Array notation is stripped before matching, so `interval[]` / `ltree[]` are covered too.

## Why string

Both types have well-defined textual representations and no natural JSON-schema numeric/object form (`interval` can't satisfy RFC3339, the same reasoning already applied to `time`/`money`).

## Tests

- New `tests/unit/test_discovery_utils.py` covering `ltree`, `interval`, primary-key (non-nullable) handling for both, and array-notation stripping.
- `pytest tests/unit/test_discovery_utils.py` → 5 passed (Python 3.12).
- `pylint --rcfile .pylintrc tap_postgres/discovery_utils.py` → 10.00/10.

## Notes

- Bumps `pipelinewise-tap-postgres` to `2.3.0` with a CHANGELOG entry (marked unreleased — happy to adjust the version/date to your release flow).
- Addressed review feedback: collapsed the two type checks into a single `data_type in {...}` membership test, and added a primary-key `interval` test.
- Small, additive change, but since Postgres is a connector you use at Wise I'm glad to discuss via an issue first or adjust anything you'd prefer.

### DIFF
--- a/singer-connectors/tap-postgres/CHANGELOG.md
+++ b/singer-connectors/tap-postgres/CHANGELOG.md
@@ -1,3 +1,8 @@
+2.3.0 (unreleased)
+-------------------
+**Changes**
+- DISCOVERY: Map `ltree` and `interval` columns to a string schema instead of leaving them untyped.
+
 2.2.0 (2026-04-07)
 -------------------
 **Changes**
--- a/singer-connectors/tap-postgres/setup.py
+++ b/singer-connectors/tap-postgres/setup.py
@@ -6,7 +6,7 @@ with open('README.md') as f:
     long_description = f.read()
 
 setup(name='pipelinewise-tap-postgres',
-      version='2.2.0',
+      version='2.3.0',
       description='Singer.io tap for extracting data from PostgresSQL - PipelineWise compatible',
       long_description=long_description,
       long_description_content_type='text/markdown',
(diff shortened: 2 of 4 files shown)


cameron-process-st:
Thanks for the review. Addressed the feedback:
- Collapsed the `interval`/`ltree` branches into a single `if data_type in {'interval', 'ltree'}:` membership check.
- Added `test_interval_primary_key_is_not_nullable` to match the ltree PK test.

Unit tests pass (5) and `pylint` is 10.00/10 on the changed module. Re the suppressed note about the "non-array datatype" docstring vs the `[]` stripping — that's pre-existing behavior I left untouched to keep this PR focused, but happy to tidy it in a follow-up if you'd like.

PR #1287, opened 2026-07-29
feat(target-snowflake): support SSE-KMS server-side encryption on S3 uploads

mareebsiddiqui:
## Context

As part of the IAM user → IAM role migration (AP-3133), we're replacing client-side encryption with SSE-KMS for the Snowflake ETL stage. This change adds optional SSE-KMS support to the S3 upload path so that objects are encrypted at rest using a specified KMS key, without requiring client-side encryption keys in the config.

When `s3_server_side_encryption_kms_key_id` is present in the target config, uploads include `ServerSideEncryption: aws:kms` and `SSEKMSKeyId` headers. When absent, behavior is unchanged — fully backwards-compatible.

## Types of changes

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


louis-pie:
This PR cannot be merged as it only deals with the singer target-snowflake. PipelineWise's internal FastSync will ignore the new KMS settings.

pipelinewise/fastsync/commons/target_snowflake.py
```diff
@@ -144,10 +144,14 @@ def upload_to_s3(self, file, tmp_dir=None):
...
+        # Upload to S3 without client-side encrypting
         else:
-            extra_args = {'ACL': s3_acl} if s3_acl else None
-            self.s3.upload_file(file, bucket, s3_key, ExtraArgs=extra_args)
+            extra_args = {'ACL': s3_acl} if s3_acl else {}
+            kms_key_id = self.connection_config.get('s3_server_side_encryption_kms_key_id', '')
+            if kms_key_id:
+                extra_args['ServerSideEncryption'] = 'aws:kms'
+                extra_args['SSEKMSKeyId'] = kms_key_id
+            self.s3.upload_file(file, bucket, s3_key, ExtraArgs=extra_args or None)
```
louis-pie, line 185: Thanks, this now covers the shared FastSync uploader, so both FullSync and PartialSync will pick up the KMS headers. Could you add equivalent unit coverage in `tests/units/fastsync/commons/test_fastsync_target_snowflake.py` for no key, KMS, KMS+ACL, and client-side-encryption precedence? The current tests only exercise the Singer upload client, so this new FastSync branch is untested.

louis-pie (commented):
Thanks, the upload implementation now covers Singer and the shared FastSync path. I have three follow-up points to close out the operational and test coverage.

docs/connectors/targets/snowflake.rst
```diff
@@ -130,6 +130,11 @@ Example YAML for target-snowflake:
       # The same master key has to be added to the external stage object created in snowflake
       #client_side_encryption_master_key: "<MASTER_KEY>" # Plain string or vault encrypted
 
+      # Optional: Server Side Encryption (SSE-KMS)
+      # When set, uploads to S3 include SSE-KMS headers so objects are encrypted at rest
+      # using the specified KMS key. Ignored when client_side_encryption_master_key is set.
+      #s3_server_side_encryption_kms_key_id: "<KMS_KEY_ID>" # KMS key ID or ARN
```
louis-pie: Is the Snowflake stage and IAM/KMS policy change handled elsewhere under AP-3133? The page still creates the stage with `MASTER_KEY`, but SSE-KMS loads need the stage configured for `TYPE='AWS_SSE_KMS'`, plus KMS permissions for the uploader and Snowflake storage-integration identities. If this is external work, could we link it or document the prerequisite here?

singer-connectors/target-snowflake/target_snowflake/upload_clients/s3_upload_client.py
```diff
@@ -83,10 +83,14 @@ def upload_file(self, file, stream, temp_dir=None):
...
+        # Upload to S3 without client-side encrypting
         else:
-            extra_args = {'ACL': s3_acl} if s3_acl else None
-            self.s3_client.upload_file(file, bucket, s3_key, ExtraArgs=extra_args)
+            extra_args = {'ACL': s3_acl} if s3_acl else {}
+            kms_key_id = self.connection_config.get('s3_server_side_encryption_kms_key_id', '')
+            if kms_key_id:
+                extra_args['ServerSideEncryption'] = 'aws:kms'
+                extra_args['SSEKMSKeyId'] = kms_key_id
+            self.s3_client.upload_file(file, bucket, s3_key, ExtraArgs=extra_args or None)
```
louis-pie, line 124: How should `archive_load_files` behave with this setting? Both archive copy paths omit encryption parameters, so the copied object uses the destination bucket's default encryption rather than necessarily this configured key. If that is intentional, documenting the required archive-bucket default would be enough; otherwise we may need archive-specific KMS handling.

singer-connectors/target-snowflake/target_snowflake/upload_clients/s3_upload_client.py
```diff
@@ -44,6 +44,30 @@ def _create_s3_client(self, config=None):
...
+
+        if encryption_type is None or encryption_type.lower() == 'none':
+            return {}
+
+        if encryption_type.lower() != 'kms':
+            raise NotImplementedError(
+                f"Encryption type '{encryption_type}' is not supported. Expected: 'none' or 'KMS'"
+            )
+
+        encryption_args = {'ServerSideEncryption': 'aws:kms'}
```
louis-pie, line 64: Passing `ServerSideEncryption='aws:kms'` without `SSEKMSKeyId` does not use the bucket's customer-managed default key; S3 uses the AWS-managed `aws/s3` key. That contradicts the docs and can fail a bucket policy requiring the AP-3133 key. Please either require `encryption_key` for KMS (my preference), or send no SSE headers when relying on bucket defaults, and update both Singer and FastSync tests accordingly. AWS documents the behaviour here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-kms-encryption.html

singer-connectors/target-snowflake/tests/unit/test_s3_upload_client.py
```diff
@@ -0,0 +1,187 @@
+import os
+import unittest
+from unittest.mock import patch, MagicMock
+
+from target_snowflake.upload_clients.s3_upload_client import S3UploadClient
+
+
+class TestS3UploadClient(unittest.TestCase):
```
louis-pie, line 8: These target-snowflake tests are not currently run by CI. The connector unit matrix only includes tap-mysql and tap-postgres, and this connector exposes `unit_test` rather than `unit_test_cov`. Please wire target-snowflake into connector CI so the new Singer KMS path is actually executed; the green E2E jobs do not enable KMS.

docs/connectors/targets/snowflake.rst
```diff
@@ -130,6 +132,103 @@ Example YAML for target-snowflake:
...
+     - ``CopyObject`` decrypts the load file and re-encrypts the archived copy
+   * - The storage integration's IAM role, or the stage credentials
+     - ``kms:Decrypt``
+     - Snowflake decrypts the load files during ``COPY INTO``
+
+.. note::
+
+  PipelineWise sends the same ``encryption_key`` for the archived copy regardless of
+  destination, so when ``archive_load_files_s3_bucket`` names a different bucket that key
+  must be usable for it. Cross-Region buckets need their own key: an S3 KMS key must be in
```
louis-pie, line 226: This notes that a cross-Region archive needs a different regional key, but both archive paths always reuse `encryption_key` and there is no archive-specific setting. Please either add an archive key option or state explicitly that cross-Region archive buckets are unsupported with SSE-KMS.

docs/connectors/targets/snowflake.rst
```diff
@@ -130,6 +132,103 @@ Example YAML for target-snowflake:
...
+      # Optional: Server Side Encryption (SSE-KMS)
+      # Requires the stage and KMS prerequisites in the section below. Applies to load file
+      # uploads and to archived load file copies. Takes precedence over
+      # client_side_encryption_master_key: when both are set, files are not client side
+      # encrypted, so the stage must not declare AWS_CSE encryption.
+      #encryption_type: "KMS"                       # (Default: None) The type of encryption to use. Current supported options are: 'none' and 'KMS'.
+      #encryption_key: "<ENCRYPTION_KEY_ID>"        # Optional: The KMS key ID (e.g. '1234abcd-1234-1234-1234-1234abcd1234'), key ARN or alias.
+                                                    # An alias resolves in the account making the request.
+                                                    # If omitted, the S3 bucket default KMS key is used.
+                                                    # This field is ignored if 'encryption_type' is none or blank.
```
louis-pie, line 144: The config surface is still inconsistent: the PR description names the removed `s3_server_side_encryption_kms_key_id`, the canonical target Snowflake sample does not include `encryption_type` / `encryption_key`, and this says a blank `encryption_type` is ignored although the implementation raises `NotImplementedError`. Please align all three with the final behaviour.

louis-pie (commented):
The earlier FastSync, stage and archive gaps are now largely addressed. I found four remaining points to close before merge.

mareebsiddiqui:
Not needed anymore as KMS isn't required for AP-3133

PR #1288, opened 2026-07-31
[AP-2830] Add data-diff checks for source-to-target reconciliation

louis-pie:
## Context

Replication has no way to answer "did the target actually receive what the source
has?" — drift from silent FastSync coercions, missed binlog events, or partial loads
is invisible until someone notices downstream.

This adds **data-diff**: bounded aggregate reconciliation between a source table and
its PostgreSQL or Snowflake replica. Checks are declared per table in the tap YAML,
persisted as immutable versioned definitions, and run on a crontab cadence over a
settled time window. A coverage watermark (`verified_through`) advances only across a
contiguous union of passing windows, so a gap or failure is visible rather than
averaged away, and remediation preserves the original failure as evidence.

Design constraints worth flagging for review:

- **No source rows or business values are ever stored** — only aggregate metrics.
  Comparison is `COUNT`/`MIN`/`MAX`-style aggregates plus an optional checksum.
- **Source safety is explicit.** Read-only transactions, a per-query
  `statement_timeout`, and a preflight that refuses a check whose source table is
  large and has no usable index leading with the timestamp column. The preflight is
  persisted *before* either aggregate query runs, so a `PASS` stays auditable.
- **The backend database is optional and isolated.** Without a `backend_db` block,
  `import_config` skips every `data_diff` block — no existing config changes
  behaviour. Replication never reads the backend, so an outage there pauses
  reconciliation only.
- **Migrations run as a separate `ddl_user`** that owns the schema and grants the
  application user DML only, so a compromised pipeline cannot alter the schema.

Four routes are supported: {PostgreSQL, MySQL/MariaDB} sources × {PostgreSQL,
Snowflake} targets.

`row_checksum` ships as **experimental** — it is probabilistic, identifies a window
rather than individual rows, and cannot be trusted on JSON/VARIANT (non-deterministic
key ordering) or FLOAT (IEEE 754) columns. This is documented at the point of use.

Alongside the feature, this fixes several defects in existing master-branch code that
the work surfaced — most consequentially, MySQL/MariaDB replication persisting invalid
dates instead of NULL, and the VictorOps handler failing to send **any** alert that
carried an exception (i.e. every tap and fast-sync failure). See CHANGELOG for the
full list.

[Add data-diff checks - PipelineWise documentation.pdf](https://github.com/user-attachments/files/30590952/AP-2830.Add.data-diff.checks.-.PipelineWise.documentation.pdf)

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) [optional]
- [x] New feature (non-breaking change which adds functionality) [optional]
- [x] Documentation Update (if none of the other choices apply) [optional]

## Checklist

[AP-2830]: https://transferwise.atlassian.net/browse/AP-2830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pipelinewise/backend_db/migrations/versions/001_initial_data_diff_schema.py
```diff
@@ -0,0 +1,260 @@
...
+                CHECK (trigger IN ('SCHEDULED', 'MANUAL', 'RETRY', 'REMEDIATION')),
+            status TEXT NOT NULL CHECK (status IN ('RUNNING', 'PASS', 'FAIL', 'ERROR')),
+            rerun_of_run_id UUID REFERENCES {SCHEMA}.dd_runs(run_id) ON DELETE RESTRICT,
+            remediation_reference TEXT,
+            preflight_id UUID REFERENCES {SCHEMA}.dd_preflights(preflight_id) ON DELETE RESTRICT,
+            started_at TIMESTAMPTZ NOT NULL,
+            finished_at TIMESTAMPTZ,
+            error TEXT,
+            CHECK (window_start < window_end),
+            CHECK (num_nonnulls(dd_check_id) = 1),
```
amofakhar: If I'm correct, num_nonnulls requires at least 2 arguments and passing a single argument raises an error. maybe can change it to CHECK (dd_check_id IS NOT NULL)

pipelinewise/backend_db/migrations/versions/001_initial_data_diff_schema.py
```diff
@@ -0,0 +1,260 @@
...
+        op.execute(f"GRANT SELECT ON {SCHEMA}.{view} TO {role}")
+
+    # dd_coverage_events.event_sequence is a BIGSERIAL: inserting needs the
+    # sequence, not just the table.
+    op.execute(
+        f"GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA {SCHEMA} TO {role}"
+    )
+
+    # alembic_version is read on every connection that checks the schema version.
+    op.execute(f"GRANT SELECT ON {SCHEMA}.alembic_version TO {role}")
```
amofakhar, line 293: When Alembic runs revision 001 for the first time on a fresh database, it executes upgrade() before it creates the alembic_version and it will throw an error.


```suggestion
    op.execute(f"""
    DO $$ 
    BEGIN 
        IF to_regclass('{SCHEMA}.alembic_version') IS NOT NULL THEN
            EXECUTE 'GRANT SELECT ON {SCHEMA}.alembic_version TO {role}';
        END IF;
    END $$;
""")
```

pipelinewise/backend_db/migrations/versions/001_initial_data_diff_schema.py
louis-pie: Hey, this code works, but maybe a little confusing. I’ll make dd_check_id explicitly NOT NULL and remove the check.

pipelinewise/backend_db/migrations/versions/001_initial_data_diff_schema.py
louis-pie: On a fresh database, Alembic creates alembic_version before invoking revision upgrade(), so the table exists when this grant runs.

We also have an E2E to verifies it works on a clean backend.[https://github.com/transferwise/pipelinewise/blob/c3ad5b5f635884ea327eaf41f3a28be542a3c903/tests/end_to_end/helpers/env.py#L573-L585](https://github.com/transferwise/pipelinewise/blob/c3ad5b5f635884ea327eaf41f3a28be542a3c903/tests/end_to_end/helpers/env.py#L573-L585)

Is there another edge case I am not thinking of?

pipelinewise/backend_db/migrations/env.py
```diff
@@ -0,0 +1,28 @@
...
+def run_migrations_online():
+    """Run migrations using the connection URL set by BackendDatabase."""
+    connectable = engine_from_config(
+        context.config.get_section(context.config.config_ini_section, {}),
+        prefix="sqlalchemy.",
+        poolclass=pool.NullPool,
+    )
+
+    with connectable.connect() as connection:
+        context.configure(
```
amofakhar, line 19: If schema is changed from public to something else, Alembic will fail to inspect or track versioning.
```suggestion
        context.configure(
                connection=connection,
                target_metadata=None,
                compare_type=True,
                include_schemas=True,
                version_table_schema=context.config.get_main_option("version_table_schema", None)
            
```

pipelinewise/backend_db/database.py
```diff
@@ -0,0 +1,157 @@
...
+                f"postgresql+psycopg2://{quote_plus(conn['user'])}:{quote_plus(conn['password'])}"
+                f"@{conn['host']}:{conn.get('port', 5432)}/{conn['dbname']}"
+            )
+            query = {
+                key: conn[key]
+                for key in ("options", "sslmode")
+                if conn.get(key)
+            }
+            if query:
+                url += "?" + "&".join(
```
amofakhar: Options strings in PG often contain spaces or specific flags. `quote_plus` encodes spaces as +, which SQLAlchemy query parser sometimes fail to unpack correctly for the options parameter.

use standard url lib.parse.urlencode with quote_via=quote or construct a SQLAlchemy URL object by using sqlalchemy.engine.URL.create()

pipelinewise/backend_db/migrations/env.py
louis-pie: True, will be more explicit about the destination schema and update E2E tests

pipelinewise/data_diff/comparison.py
```diff
@@ -0,0 +1,181 @@
...
+        source_descriptors.append({**source_column, **common})
+        target_descriptors.append({**target_column, **common})
+    return source_descriptors, target_descriptors
+
+
+def metric_passes(check_type: str, source_value, target_value) -> bool:
+    """Apply equality checks and side-specific key-integrity invariants."""
+    if check_type in ("null_key_count", "duplicate_key_count"):
+        try:
+            return Decimal(str(source_value)) == 0 and Decimal(str(target_value)) == 0
```
amofakhar, line 176: Is it possible if the `source_value` or `target_value` be `None`? if yes then str(None) = "None" => Decimal("None") raises an error.

pipelinewise/backend_db/database.py
louis-pie: Adding some safeguards and new tests

pipelinewise/data_diff/comparison.py
louis-pie: I think this one is not correct.

The count-based checks should always return numeric values because they use COUNT. If an adapter unexpectedly returns None, Decimal("None") raises decimal.InvalidOperation, which inherits from ArithmeticError and is already caught here, so the check returns False rather than raising. MIN and MAX can return None for empty windows, but those checks use direct equality rather than Decimal.

pipelinewise/backend_db/database.py
louis-pie: Safeguards added

pipelinewise/backend_db/migrations/env.py
louis-pie: updated schema explicitness and tests

pipelinewise/backend_db/migrations/versions/001_initial_data_diff_schema.py
louis-pie: modified schema slightly to remove ambiguity

pipelinewise/data_diff/adapters.py
```diff
@@ -0,0 +1,594 @@
...
+        )
+
+    def execute_metrics(self, sql, params, checks):
+        import snowflake.connector  # pylint: disable=import-outside-toplevel
+
+        started = perf_counter()
+        with self.connection.cursor(snowflake.connector.DictCursor) as cursor:
+            cursor.execute(sql, params)
+            row = cursor.fetchone()
+        values_by_name = {str(key).lower(): value for key, value in row.items()}
```
amofakhar, line 552: what if row be None? it will raise and exception here!

louis-pie: Good catch. The aggregate query should always return one row, even for an empty window.
I’ll add the same `DataDiffExecutionError` guard and a regression test.

pipelinewise/data_diff/comparison.py
```diff
@@ -0,0 +1,181 @@
...
+        target_family = type_family(target_column)
+        if not _schema_families_compatible(source_family, target_family):
+            raise UnsupportedComparisonError(
+                f"row_checksum column '{logical_name}' has incompatible source "
+                f"and target types ({source_family}, {target_family})"
+            )
+
+        checksum_kind = source_family
+        scale = 0
+        if source_family == "date" and target_family == "timestamp":
```
amofakhar: in line #165, there is :

```Python
if checksum_kind == "date":
            checksum_kind = "timestamp"
```

wha is the use to have here
```Python
... and target_family == "timestamp"
```

finally `checksum_kind` will be `timestap` if checksum_kind == source_family == "date"

pipelinewise/data_diff/coverage.py
```diff
@@ -0,0 +1,153 @@
...
+
+    if data_checks_enabled:
+        passing = sorted(
+            (run for run in effective if run["status"] == "PASS"),
+            key=lambda item: (item["window_start"], item["window_end"]),
+        )
+        for run in passing:
+            if run["window_start"] > cursor:
+                break
+            if run["window_end"] > cursor:
```
amofakhar, line 43: Found By AI:

When iterating over passing runs, if a run's window_start is strictly less than cursor, its interval is merged correctly:
```Python
if run["window_start"] > cursor:
    break
if run["window_end"] > cursor:
    cursor = run["window_end"]
```

However, if a run completely overlaps or is out-of-order in terms of window_end (e.g., Run A: 0..10, Run B: 2..5), Run B will have run["window_end"] <= cursor, so cursor remains 10. If Run C is 6..12, Run C will correctly advance cursor to 12.

The issue: If the runs list contains overlapping or out-of-order windows that aren't strictly sorted by window_end alongside window_start, sorting solely by (window_start, window_end) causes a gap check failure if a gap occurs within a run that was skipped because its start was valid.

pipelinewise/data_diff/coverage.py
```diff
@@ -0,0 +1,153 @@
...
+def calculate_coverage(runs: list, *, data_checks_enabled: bool = True) -> dict:
+    """Calculate the conservative contiguous interval covered by effective PASS runs."""
+    effective = effective_attempts(runs)
+    if not effective:
+        return {}
+
+    coverage_start = min(run["window_start"] for run in effective)
+    max_observed_end = max(run["window_end"] for run in effective)
+    cursor = coverage_start
+    blocking_runs = sorted(
```
amofakhar, line 29: Found by AI:

Setting barrier to blocking_runs[0]["window_start"] assumes the earliest failing run is the primary barrier. If there is a failing run far in the future (e.g., window_start = 100) and valid passing runs covering 0..50, the cursor loop checks cursor >= barrier on every pass. While cursor will stop at 100, the report will assign blocking_run = blocking_runs[0] even if cursor was blocked earlier by a gap at 50.

```Python
blocking_run = None
if barrier is not None and barrier <= cursor:
    blocking_run = blocking_runs[0]
```

If a gap stopped cursor at 50, but barrier is 100, barrier <= cursor evaluates to 100 <= 50 (False). The status becomes BLOCKED, but blocking_run remains None. This creates a mismatch where the reason says "No successful run covers the next timestamp interval" even though a failed run exists in the pipeline.

pipelinewise/data_diff/coverage.py
```diff
@@ -0,0 +1,153 @@
+"""Derive contiguous timestamp coverage from immutable check attempts."""
+
+
+TERMINAL_STATUSES = {"PASS", "FAIL", "ERROR"}
+
+
+def effective_attempts(runs: list) -> list:
```
amofakhar: it is an internal method, isn't it?

pipelinewise/data_diff/engine.py
```diff
@@ -0,0 +1,565 @@
...
+        }
+        ssl_context, ssl_temp_dir = _mysql_ssl_context(connection_config)
+        # Opportunistic TLS, matching tap-mysql and FastSync. Without it a server
+        # requiring encryption rejects us with a misleading "Access denied".
+        connection_args["ssl"] = ssl_context if ssl_context else {"": True}
+        try:
+            connection = pymysql.connect(**connection_args)
+        finally:
+            if ssl_temp_dir:
+                ssl_temp_dir.cleanup()
```
amofakhar, line 241: connection context Is still using and if a negotiation be needed or during the session needs to read the cert file, it will fail!

pipelinewise/data_diff/coverage.py
louis-pie: indeed, will address

pipelinewise/data_diff/comparison.py
louis-pie: doesn't impact functionality, but easy enough to make clearer. will address

pipelinewise/data_diff/coverage.py
louis-pie: I don't think this is valid

Sorting intervals by start and extending the cursor to the maximum reachable end is correct. A nested interval cannot hide a gap from a later interval.

pipelinewise/data_diff/coverage.py
louis-pie: I think it might be mistaken

If successful coverage ends at 50 and a failed run starts at 100, the current boundary is the uncovered 50..100 gap. Reporting no blocking_run_id and “No successful run covers the next timestamp interval” is correct.

pipelinewise/data_diff/engine.py
louis-pie: Doesn't seem valid

This is the MySQL connection path, not Snowflake. _mysql_ssl_context() loads the certificate and key into an SSLContext before returning. pymysql.connect() synchronously wraps and authenticates the socket before returning, and only then are the temporary files removed.

amofakhar (approved):

PR #1289, opened 2026-08-05
[AP-2830] Fix MySQL and MariaDB date-time replication

louis-pie:
## Context

This fixes a bug surfaced during https://github.com/transferwise/pipelinewise/pull/1288.
This fix needs to be merged before creating a new Release `0.78.0`

The MySQL/MariaDB tap treated every string value with a `date-time` schema format as invalid and converted it to `null`. In LOG_BASED replication this also removed valid ISO timestamps such as `_sdc_deleted_at`, breaking soft-delete metadata.

This PR:

- Preserves valid ISO date-time strings while converting malformed MySQL/MariaDB `date`, `datetime`, and `timestamp` values to `null` across FULL_TABLE, INCREMENTAL, and LOG_BASED replication.
- Builds binlog stream metadata before emitting schema messages so automatic properties such as `_sdc_deleted_at` are included.
- Adds unit coverage for valid and malformed timestamps, automatic schema properties, and timezone isolation.
- Adds MariaDB-to-Snowflake E2E coverage for both hard and soft deletes.
- Moves the tap-mysql unit suite to Pytest and adds Python 3.12 CI coverage gates for tap-mysql and tap-postgres.
- Records the tap-mysql fixes in the PipelineWise release notes.
- Refreshes repository guidance for connector environments, CI validation, and safe E2E execution.

## Types of changes

*Put an `x` in the boxes that apply*

- [x] Bugfix (non-breaking change which fixes an issue) [optional]

## Checklist


singer-connectors/tap-mysql/tap_mysql/sync_strategies/common.py
```diff
@@ -5,13 +5,18 @@
...
 
+from pymysql.converters import convert_datetime
 from singer import metadata, utils, metrics
 
 from tap_mysql.stream_utils import get_key_properties
 
 LOGGER = singer.get_logger('tap_mysql')
 
 
+def is_invalid_mysql_datetime(value):
```
amofakhar, line 15: I believe there is no unit test for this method!

it returns False for below values which is not correct!

None
"INVALID_MYSQL_DATETIME".encode("utf-8")
4
[datetime.now(), datetime.now()]
["Foo", "bar"]

louis-pie: Good point. It is exercised through both row_to_singer_record paths, butthe helper itself does not have a focused unit test.

Will add the test

singer-connectors/tap-mysql/tests/unit/sync_strategies/test_common.py
```diff
@@ -33,3 +33,48 @@ def test_row_to_singer_record(self):
...
         assert message.time_extracted is not None
+
+    def test_row_to_singer_record_only_nulls_invalid_datetime_strings(self):
+        datetime_values = {
+            'valid_datetime': '2024-02-29T12:00:00+00:00',
+            'zero_datetime': '0000-00-00 00:00:00',
+            'zero_year': '0000-01-01 00:00:00',
+            'invalid_month': '2024-13-01 12:00:00',
+            'zero_day': '2024-05-00 12:00:00',
+            'invalid_day': '2024-02-30 12:00:00',
```
amofakhar, line 56: '2024-02-29T12:00:00+99:00' is an invalid value, it is missed in this test and also it does not return it as None.
the same for  '0000-00-00 00:00:00'.encode('utf-8')

CHANGELOG.md
```diff
@@ -1,28 +1,31 @@
-0.78.0 (2026-07-31)
+0.78.0 (2026-08-05)
 -------------------
 
 **Connector updates**
 
 - `pipelinewise-tap-mixpanel` from `1.7.1` to `1.7.2`
     - Restore `mp_reserved_insert_id` as the export stream key property so newly discovered taps pass primary-key validation
     - Respect `denest_properties=false` during discovery so dynamic Mixpanel properties remain nested
+- `pipelinewise-tap-mysql` from `1.6.0` to `1.6.1`
```
amofakhar: I think, it's better not to bump version for singer connectors here.We can just include the code changes.
there are two versions of connectors like pipelinewise-tap-mysql and pipelinewise-tap-mixpanel:

1. The PyPI package (which we decided to stop releasing).
2. The code embedded directly inside the Pipelinewise codebase.

In other words, there won't be a version 1.6.1 for pipelinewise-tap-mysql; we are only updating the internal Pipelinewise code here.

singer-connectors/tap-mysql/setup.py
```diff
@@ -6,7 +6,7 @@
     long_description = fh.read()
 
 setup(name='pipelinewise-tap-mysql',
-      version='1.6.0',
+      version='1.6.1',
```
amofakhar: there won't be any release for `pipelinewise-tap-mysql` !

CHANGELOG.md
louis-pie: Makes sense, will address

singer-connectors/tap-mysql/setup.py
louis-pie: Makes sense, will address

singer-connectors/tap-mysql/tests/unit/sync_strategies/test_common.py
louis-pie: makes sense, will address

amofakhar (approved):

PR #1329, opened 2026-08-24
Snowflake Iceberg v3 support for RDBMS sources

louis-pie:
## Context

PipelineWise can create Snowflake-managed Iceberg tables through Singer, but the
existing RDBMS FastSync and PartialSync publication lifecycle was native-table
specific. AP-2520 adds a fail-closed managed Iceberg v3 lifecycle for MariaDB,
MySQL, and PostgreSQL sources, including crash recovery, state ordering, and a
controlled native-to-Iceberg conversion path.

See the complete [0.81.0 CHANGELOG entry](https://github.com/transferwise/pipelinewise/blob/a1330d3a09a543cca5a571b35de674ba0d2fe057/CHANGELOG.md#0810-2026-08-24).

## Changes

- Select native or managed Iceberg v3 tables per tap and validate the source,
  target, replication method, table format, and Iceberg version before dispatch.
- Add managed Iceberg v3 FullSync and PartialSync publication for MariaDB,
  MySQL, and PostgreSQL, with table-level copy-on-write, wide physical types,
  staged-content verification, and state handoff only after finalization.
- Add target-scoped recovery coordination, strict typed manifests, query-history
  reconciliation, and a configurable active-query polling budget.
- Add the PipelineWise-owned `copy_native_to_iceberg` utility with guarded
  metadata preservation, exact content verification, rollback, and retry.
- Align target-snowflake and FastSync type contracts, MariaDB JSON aliases,
  PostgreSQL hstore, string widths, timestamps, binary values, and approximate
  numerics for managed Iceberg v3.
- Expand unit, connector, credentialed E2E, CI, configuration, operational
  documentation, and release notes for the new lifecycle.

## Type of change

- [x] Bug fix
- [x] New functionality
- [x] Maintenance, dependency, or CI
- [x] Documentation
- [x] Breaking change

## Compatibility and operations

- Managed Iceberg v2 and the deprecated target-level `iceberg_create` setting
  are removed. Managed Iceberg taps must select `target_table_format: iceberg`,
  use integer `iceberg_version: 3`, enable `hard_delete`, and remain unflattened
  on FastSync-capable RDBMS routes.
- PipelineWise must be the sole writer. Existing PipelineWise-created managed
  v3 targets must have table-level
  `ICEBERG_MERGE_ON_READ_BEHAVIOR = 'DISABLED'` and all string columns must be
  `VARCHAR(134217728)` before PipelineWise writes to them.
- Recovery depends on both target-scoped FastSync recovery files. Operators must
  preserve them and retry the same command unchanged when publication outcome is
  ambiguous.
- Native-to-Iceberg cutover requires a controlled reader-and-writer outage. The
  converter retains the native table as the rollback point and rejects unsupported
  dependencies or metadata before copying.
- Release 0.80.0 and its tag were deleted. Every still-supported 0.80.0 change is
  folded into 0.81.0.

## Validation

| Command or check | Result |
|---|---|
| `ruff check pipelinewise tests` | Passed |
| `pylint pipelinewise tests` | Passed, 10.00/10 |
| `flake8 pipelinewise --count --select=E9,F63,F7,F82 --show-source --statistics` | Passed, 0 findings |
| `flake8 pipelinewise --count --max-complexity=15 --max-line-length=120 --statistics` | Passed, 0 findings |
| `pytest --cov=pipelinewise --cov-fail-under=77 -v tests/units` | Passed: 1,252 tests and 143 subtests; 65 warnings; 86% coverage |
| `pipelinewise validate --dir dev-project/pipelinewise-config` | Passed |
| `cd singer-connectors/target-snowflake && make unit_test` | Passed: 123 tests and 61 subtests; 81.05% coverage |
| `cd singer-connectors/tap-mysql && make unit_test_cov` | Passed: 49 tests; 56.45% coverage |
| `cd docs && make check` | Passed: 6 reference tests and strict 45-page Sphinx build |
| Serial credentialed Snowflake publisher, MySQL, PostgreSQL, MariaDB, converter, and target-snowflake v3 runs | Passed earlier on the branch; not rerun after the final recovery/configuration hardening |
| `git diff --check` | Passed |

The target-snowflake repository-wide `make pylint` target remains non-green
because its legacy `pylintrc` is incompatible with the installed Pylint 4 and
reports pre-existing findings. Focused changed-module Pylint checks passed.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md).
- [x] This pull request is focused and can be reviewed and reverted
      independently.
- [x] Behavioural changes have regression tests, or I explained why tests are not
      applicable.
- [x] Relevant documentation and changelog entries are updated, or I explained
      why they are not applicable.
- [x] Applicable local checks pass; failures, skips, and unavailable checks are
      documented above.
- [x] The change contains no credentials, private keys, production identifiers,
      or source records.
- [x] Breaking and compatibility impacts are identified above.


pipelinewise/cli/__init__.py
```diff
@@ -186,9 +186,32 @@ def _validate_command_specific_arguments(args):
     if args.command == 'partial_sync_table':
         _validate_partial_sync_arguments(args)
 
+    if args.command == 'copy_native_to_iceberg':
+        _validate_iceberg_copy_arguments(args)
+
     _validate_data_diff_arguments(args)
 
 
+def _validate_iceberg_copy_arguments(args):
```
amofakhar, line 195: It accepts empty string "" or whitespace!

pipelinewise/cli/config.py
```diff
@@ -12,15 +12,24 @@
 from pipelinewise.utils import safe_column_name
 from . import utils
 from .errors import InvalidConfigException
+from .fastsync_capabilities import (
+    TABLE_FORMAT_ICEBERG as ICEBERG_TABLE_FORMAT,
```
amofakhar: Importing TABLE_FORMAT_ICEBERG as ICEBERG_TABLE_FORMAT and TABLE_FORMAT_NATIVE as NATIVE_TABLE_FORMAT only to re-assign it inside the class as TABLE_FORMAT_ICEBERG = ICEBERG_TABLE_FORMAT creates a confusing and circular translation layer.
Better to use the imported constants (NATIVE_TABLE_FORMAT and ICEBERG_TABLE_FORMAT) directly throughout the class methods instead of aliasing them.

pipelinewise/cli/__init__.py
louis-pie: Makes sense, will address

pipelinewise/cli/config.py
louis-pie: Makes sense, will address

pipelinewise/fastsync/commons/rdbms_source.py
```diff
@@ -0,0 +1,154 @@
...
+        snowflake_types = None
+        if iceberg_requested:
+            snowflake_types = source.map_column_types_to_target(table)
+            validate_types(snowflake_types)
+
+        file_parts, size_bytes = inspect_export()
+        if snowflake_types is None:
+            snowflake_types = source.map_column_types_to_target(table)
+        source.close_connection()
+        return snowflake_types, file_parts, size_bytes
```
amofakhar: This can be simpler and cleaner:

```suggestion
        snowflake_types = source.map_column_types_to_target(table)
        if iceberg_requested:
            validate_types(snowflake_types)

        file_parts, size_bytes = inspect_export()
        source.close_connection()
        return snowflake_types, file_parts, size_bytes
```

pipelinewise/fastsync/commons/rdbms_to_snowflake.py
```diff
@@ -0,0 +1,423 @@
...
+    run.route_utils.finalize_snowflake_fullsync(
+        run.snowflake,
+        run.s3_keys,
+        run.args.target.get('s3_bucket'),
+        run.args.target,
+        run.target_schema,
+        run.table,
+        publication_error=publication_error,
+    )
+    run.s3_keys = []
```
amofakhar, line 418: I think this should be cleared after checking `publication_error` because if there is a publication error, then error will be raised and in _full_failure_result there is `getattr(exc, 's3_keys', run.s3_keys)` and because run.s3_keys is already emptied, any S3 object that were not cleaned up during `finalize_snowflake_fullsync` will leak.

pipelinewise/fastsync/commons/rdbms_to_snowflake.py
```diff
@@ -0,0 +1,423 @@
...
+        logger=logger,
+        route_utils=route_utils,
+        iceberg_version=iceberg_version,
+        iceberg_requested=iceberg_version is not None,
+    )
+
+    try:
+        if _prepare_full_run(run):
+            return True
+        _export_full_source(run)
```
amofakhar, line 70: In _export_full_source, copy_table exports chunked files to run.filepath. If _upload_full_export fails , _full_failure_result handles S3 staging cleanup, but the temporary files created under run.filepath on disk may remain there until temp_dir is purged.

Better to have a cleanup for run.file_parts in `finally`

pipelinewise/fastsync/commons/snowflake_iceberg.py
```diff
@@ -0,0 +1,712 @@
...
+        """Persist deterministic S3 keys before the first upload can begin."""
+        if attempt.phase != PHASE_PREPARED:
+            raise RecoveryManifestError(
+                f'Cannot plan Iceberg uploads in phase {attempt.phase}'
+            )
+        planned_s3_keys = list(s3_keys)
+        if any(not isinstance(s3_key, str) or not s3_key for s3_key in planned_s3_keys):
+            raise RecoveryManifestError('Iceberg staging S3 keys must be non-empty strings')
+        if len(set(planned_s3_keys)) != len(planned_s3_keys):
+            raise RecoveryManifestError('Iceberg staging S3 keys must be unique')
```
amofakhar: this line (any ...) iterates through the list and the second line below, it is iterates through it again.by `set(...)` . not good from performance perspective.


```suggestion
seen = set()
for key in planned_s3_keys:
    if not isinstance(key, str) or not key:
        raise RecoveryManifestError('Iceberg staging S3 keys must be non-empty strings')
    if key in seen:
        raise RecoveryManifestError(f'Duplicate S3 key found: {key}')
    seen.add(key)
```

pipelinewise/fastsync/commons/snowflake_iceberg_converter.py
```diff
@@ -0,0 +1,968 @@
...
+        self.evidence_service = SnowflakeConversionEvidenceService(self)
+        self._query_tag_context = {}
+        self.finalization_validator = SnowflakeConversionFinalizationValidator(
+            self._query,
+            self._inspect_source,
+            self._assert_equal_contents,
+            self._manual_recovery_error,
+        )
+
+    def _content_evidence(self, table, columns, project):
```
amofakhar: I could not find any usage of this method!

pipelinewise/fastsync/commons/snowflake_iceberg_inspection.py
```diff
@@ -0,0 +1,199 @@
...
+                f'Snowflake returned multiple exact matches for {target.quoted}'
+            )
+        return exact_rows[0]
+
+    def inspect_table(self, target: SnowflakeObjectName) -> SnowflakeTableSnapshot:
+        """Inspect the exact target format, schema, key, and object identity."""
+        table_row = self.discover_table_row(target)
+        if table_row is None:
+            return SnowflakeTableSnapshot(TABLE_FORMAT_MISSING, None, None)
+        table_format = self.discover_table_format(target.schema, target.table)
```
amofakhar: `discover_table_format` is called immediately after `discover_table_row` . inside `discover_table_format`, `discover_table_row` is called again. it is better to update  `discover_table_format` and pass table rows to it.
it is also called inside `snowflake_iceberg.py`

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
```diff
@@ -0,0 +1,407 @@
...
+        raise _invalid_payload()
+    if 'rollback_required' in value and not isinstance(
+        value['rollback_required'],
+        bool,
+    ):
+        raise _invalid_payload()
+
+
+def _field(value: Dict[str, Any], name: str) -> Any:
+    return deepcopy(value.get(name))
```
amofakhar, line 148: `value.get(name)` does not return a dictionary. does it? 
if it is returning only string or bool, or None, ... then using deep copy is unnecessary and just causes CPU wasting.

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+def _extensions(value: Dict[str, Any], known_fields: FrozenSet[str]) -> Dict[str, Any]:
+    return {
+        key: deepcopy(item)
```
amofakhar, line 153: same comment as `_field()`

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+
+def _context(
+    present_fields: FrozenSet[str],
+    values: Dict[str, Any],
+    extensions: Dict[str, Any],
+) -> Dict[str, Any]:
+    context = deepcopy(extensions)
+    context.update({name: deepcopy(values[name]) for name in present_fields})
```
amofakhar, line 165: same comment as _field()

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
louis-pie: Extensions are deliberately forward-compatible and can contain arbitrary nested dictionaries and lists. Without the copy, a caller changing an extension through the compatibility projection could also change the typed recovery payload. The nested-mutation test covers this exact boundary, so I have kept the deepcopy here.

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
louis-pie: It can return dictionaries here. staging_config and replacement_metadata are nested dictionaries, and serialized boundary values can be dictionary-shaped too. The payload dataclasses are frozen, but values inside them are still mutable, so this copy stops callers from changing authoritative recovery state. There is a unit test that deliberately mutates the exposed projection and checks the typed payload stays unchanged, so I have kept this one.

pipelinewise/fastsync/commons/rdbms_to_snowflake.py
louis-pie: Had a closer look at this one and I think the current order is right. finalize_snowflake_fullsync raises if grants, S3 cleanup, or temp-table cleanup fail, so these fields are only cleared after all finalization work succeeds. If cleanup fails, control never reaches these lines and _full_failure_result still gets the original keys and temp-table flag. Keeping them after a successful finalizer would only make us repeat the same cleanup when the captured SWAP error is raised, so I have left this as-is.

pipelinewise/fastsync/commons/snowflake_iceberg_inspection.py
louis-pie: Good spot. inspect_table now reuses the exact SHOW TABLES row it already fetched. The public discover_table_format API still does its own lookup, but inspection no longer repeats it. Fixed in 14a6bb6a.

pipelinewise/fastsync/commons/snowflake_iceberg.py
louis-pie: Yep, fair point. I changed this to one pass with a seen set. I kept the existing generic errors and validation priority, so a later invalid key still wins over an earlier duplicate rather than exposing a staging key in the error. Added coverage too. Fixed in 14a6bb6a.

pipelinewise/fastsync/commons/snowflake_iceberg_manifest.py
louis-pie: _context returns a mutable compatibility projection, while values and extensions hold the authoritative typed recovery payload. The nested values need to be detached or changing the returned context would also change recovery state behind the frozen dataclass. That isolation is covered by test_nested_context_projection_cannot_mutate_typed_payload, so I have kept the deepcopy here.

pipelinewise/fastsync/commons/rdbms_to_snowflake.py
louis-pie: Good catch. These files are not reused on retry, so I added best-effort cleanup for the exported parts in the FullSync finally block. Cleanup errors are logged without replacing the original result, and the MySQL/PostgreSQL upload-failure coverage now expects the files to be gone. Fixed in 14a6bb6a.

pipelinewise/fastsync/commons/snowflake_iceberg_converter.py
louis-pie: Yep, you are right, this wrapper had no call sites. Removed in 14a6bb6a.

pipelinewise/fastsync/commons/snowflake_iceberg_parameters.py
```diff
@@ -0,0 +1,78 @@
...
+        MANAGED_ICEBERG_V3_SPEC,
+    )
+
+
+def _assert_managed_copy_on_write_parameter(
+    rows: Sequence[Dict[str, Any]],
+    target: SnowflakeObjectName,
+    version_spec,
+) -> None:
+    version_spec.validate_parameter_rows(rows, target, version_spec)
```
amofakhar, line 32: Passing `version_spec` as its own argument is redundant!

pipelinewise/fastsync/commons/snowflake_iceberg_publication.py
```diff
@@ -0,0 +1,1068 @@
...
+    poll_interval_seconds: float
+    timeout_seconds: float
+    lookup_timeout_seconds: float
+
+    def __post_init__(self):
+        if (
+            self.max_attempts is not None
+            and (
+                isinstance(self.max_attempts, bool)
+                or not isinstance(self.max_attempts, int)
```
amofakhar, line 107: This is cleaner and avoids multi checks:

```suggestion
               type(self.max_attempt) is not int
```

pipelinewise/fastsync/commons/snowflake_iceberg_recovery.py
```diff
@@ -0,0 +1,219 @@
...
+def validate_required_finalization_actions(
+    attempt,
+    finalization: Optional[Dict[str, bool]] = None,
+) -> None:
+    """Require the exact completed action set before terminal handoff."""
+    completed = validate_finalization_state(
+        attempt.finalization if finalization is None else finalization
+    )
+    required = required_finalization_actions(attempt)
+    if set(completed) != set(required):
```
amofakhar: `required` is already `frozenset`. calling set(required) is redundant

pipelinewise/fastsync/commons/snowflake_iceberg_validation.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+def _value(row: Dict[str, Any], name: str, default=None):
+    for key in (name, name.upper(), name.lower()):
+        if key in row:
+            return row[key]
+    return default
+
+
+def snowflake_boolean(value: Any) -> bool:
```
amofakhar, line 215: This method is exactly same as `_boolean`!

pipelinewise/fastsync/commons/snowflake_iceberg_validation.py
```diff
@@ -0,0 +1,440 @@
...
+            'Snowflake returned invalid stream dependency metadata'
+        ) from exc
+    target = (table.database, table.schema, table.table)
+    return any(
+        (reference.database, reference.schema, reference.table) == target
+        for reference in references
+    )
+
+
+def _boolean(value: Any) -> bool:
```
amofakhar: this method is exactly same as `snowflake_boolean`!

pipelinewise/fastsync/commons/snowflake_sql_client.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+    def query(self, query, params=None, query_tag_props=None):
+        """Execute one statement and return dictionary rows when present."""
+        if self.sql_logger is not None:
+            self.sql_logger.debug('Running query: %s', query)
+        with self.open_connection(query_tag_props) as connection:
+            with connection.cursor(snowflake.connector.DictCursor) as cursor:
+                cursor.execute(query, params)
+                return cursor.fetchall() if cursor.rowcount > 0 else []
```
amofakhar: based on the query type, row count can sometimes be -1 or None even when `fetchall()` contains valid rows.


```suggestion
                return cursor.fetchall() if cursor.description  else []
```

pipelinewise/fastsync/commons/tap_mysql.py
```diff
@@ -75,6 +76,26 @@ def is_mariadb(self) -> bool:
...
+        """Return whether MariaDB JSON aliases should map to Iceberg VARIANT."""
+        iceberg_version = self.connection_config.get('iceberg_version')
+        return (
+            self.connection_config['engine'].lower() == MARIADB_ENGINE
+            and (
+                self._mariadb_json_aliases_enabled
+                or (
+                    self.connection_config.get('target_table_format') == 'iceberg'
+                    and iceberg_version == 3
+                    and not isinstance(iceberg_version, bool)
```
amofakhar: since there is `iceberg_version == 3`, no need to check later if it is not boolean!

pipelinewise/fastsync/commons/utils.py
```diff
@@ -284,21 +284,43 @@ def get_expected_s3_key(snowflake, file_part):
     return expected_key if isinstance(expected_key, str) else None
 
 
+def get_expected_s3_keys(snowflake, file_parts: List[str]) -> List[str]:
+    """Resolve and validate every deterministic staging key before upload."""
+    s3_keys = [get_expected_s3_key(snowflake, file_part) for file_part in file_parts]
+    if any(not s3_key for s3_key in s3_keys):
+        raise ValueError('Iceberg staging requires deterministic S3 keys')
+    if len(set(s3_keys)) != len(s3_keys):
+        raise ValueError('Iceberg staging S3 keys must be unique')
```
amofakhar: there is a double iteration during this validation.


```suggestion
    seen = set()
    for key in s3_keys:
          if not key:
                raise ValueError('Iceberg staging requires deterministic S3 keys')
          if key in seen:
                raise ValueError('Iceberg staging S3 keys must be unique')
          seen.add(key)
```

pipelinewise/fastsync/partialsync/rdbms_to_snowflake.py
```diff
@@ -0,0 +1,484 @@
...
+    )
+
+
+def _where_clause(column_name, start_value, end_value) -> str:
+    start_value_for_query = (
+        start_value if start_value == 'NULL' else f'\'{start_value}\''
+    )
+    where_clause_sql = f' WHERE {column_name} >= {start_value_for_query}'
+    if end_value is not None:
+        where_clause_sql += f' AND {column_name} <= \'{end_value}\''
```
amofakhar: Injection possibility!

singer-connectors/tap-mysql/tap_mysql/discover_utils.py
```diff
@@ -83,6 +84,17 @@ def is_supported_column_type(column_datatype: str) -> bool:
...
 
 
+def mariadb_json_aliases_enabled(config: Dict) -> bool:
+    """Return whether MariaDB JSON aliases should use Iceberg v3 semantics."""
+    iceberg_version = config.get('iceberg_version')
+    return (
+        str(config.get('engine', '')).lower() == 'mariadb'
+        and config.get('target_table_format') == 'iceberg'
+        and iceberg_version == 3
+        and not isinstance(iceberg_version, bool)
```
amofakhar: it is already checking `iceberg_version == 3` and no need for `not isinstance(iceberg_version, bool)`

louis-pie: Yep, same one as the FastSync helper above. Simplified it here too and kept integer v3 semantics. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_iceberg_validation.py
louis-pie: Yep, you are right. I kept snowflake_boolean as the shared helper and removed the duplicate _boolean. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_iceberg_publication.py
louis-pie: I checked this one. It is a little shorter, but it is not quite equivalent: an exact type check rejects integer subclasses, while the current contract accepts integers, explicitly excludes booleans, and still needs the positive-value check. The existing form matches the other validators, so I have left it as-is.

pipelinewise/fastsync/commons/utils.py
louis-pie: Yep, fair point. This is one pass now with a seen set, while keeping missing-key validation ahead of duplicate-key validation. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_iceberg_parameters.py
louis-pie: Had a closer look at this one. It looks redundant at the call site, but validate_parameter_rows is an unbound version-strategy callback and the selected spec is one of its declared inputs. The v3 validator reads the spec parameter, behavior, and level, and the future-version test hook relies on the same contract. Removing it would raise TypeError, so I have left this as-is.

pipelinewise/fastsync/commons/tap_mysql.py
louis-pie: Yep, fair. A boolean cannot equal 3, so the explicit boolean check was redundant. I simplified this and kept explicit integer semantics as well, so 3.0 does not sneak through. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_sql_client.py
louis-pie: Good spot. Snowflake can report an unknown row count even when result metadata exists. Both query paths now use cursor.description, with coverage for None, -1, and no result metadata. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_iceberg_recovery.py
louis-pie: Yep, fair one. required is already a frozenset, so I removed the extra conversion. Fixed in 43140043.

pipelinewise/fastsync/commons/snowflake_iceberg_validation.py
louis-pie: Agreed, same fix as the thread above. _grant_signature now uses snowflake_boolean and the duplicate helper is gone. Fixed in 43140043.

pipelinewise/fastsync/partialsync/rdbms_to_snowflake.py
louis-pie: Good catch. This affected the column and both boundary values, and the same predicate was reused for Snowflake and recovery. PartialSync now persists structured boundary data, validates and quotes source columns, binds source values through each database driver, and regenerates Snowflake SQL from typed values instead of replaying raw SQL. Fixed in 43140043.

amofakhar (approved):

